PR #20, opened 2025-12-03
feat: zod validated app types with versioning

### LINKED ISSUE #6: Add Zod typings to Swarm-UI
There are Zod typings in the library, it would be good to use Zod also for the Swarm-UI so that the key store and session data stored in localStorage can also be strictly typed and in the future versioned as well.

vojtechsimetka:
resolve #6


agazso (changes requested):
When loading the accounts I get an error (see screenshot). I think it's because the ethereum account stores the encrypted data with a `0x` prefix, but the zod schema assumes that there is no prefix.

The rest of the PR looks good though.

Screenshot From 2025-12-18 18-13-33

PR #76, opened 2025-12-15
feat: account floating menu

agazso:
This PR implements the floating menu.


swarm-ui/src/lib/components/ui/flex-item.svelte
```diff
@@ -0,0 +1 @@
+<div style="flex: 1"></div>
```
vojtechsimetka: IMHO this should be `Grow` or `Spacer` or `FlexSpacer`

PR #88, opened 2025-12-17
fix: connect keep origin appdata

### LINKED ISSUE #66: Waiting screen for Wallet / Passkey flow
what to display when users trigger the wallet or passkey flow?

we have 3 cases:

1. after account creation and before creating identity
2. after creating a new identity in an exisiting account
3. after selecting an identity from the list (and here I need to correct the Figma flow because there's no more connected identity)

Currently for 1 and 2 we just keep displaying the same screen but the button turns into a spinner button

**I would use the following screens for all cases (instead of the spinner button):**

Image

Image

https://www.figma.com/design/HFX2TQhScU0P5Azf9G1weW/SwarmID_explorations?node-id=457-4979&t=7XDAbm96cunMM1eE-1

### LINKED ISSUE #80: Improve origin management
Currently we manage origin on the connect flow in the options part of the URL. This implies that any navigation needs to keep this part in the URL. Maybe there is a better way to do this

### LINKED ISSUE #86: Create identity fails when connecting an app
Both with Ethereum and Passkey. It seems that the `sessionStore.data.account` is missing

Image

agazso:
Fixes #80 
Fixes #86 
Fixes #66

### DIFF
--- a/swarm-ui/src/lib/components/create-new-account.svelte
+++ b/swarm-ui/src/lib/components/create-new-account.svelte
@@ -1,0 +1,70 @@
+<script lang="ts">
+	import EthereumLogo from '$lib/components/ethereum-logo.svelte'
+	import PasskeyLogo from '$lib/components/passkey-logo.svelte'
+	import AuthCard from '$lib/components/auth-card.svelte'
+	import ImportSwarmIdentity from '$lib/components/import-swarm-identity.svelte'
+	import Wallet from 'carbon-icons-svelte/lib/Wallet.svelte'
+	import Horizontal from '$lib/components/ui/horizontal.svelte'
+	import Vertical from '$lib/components/ui/vertical.svelte'
+	import { goto } from '$app/navigation'
+	import routes from '$lib/routes'
+
+	function handlePasskeyClick() {
+		goto(routes.PASSKEY_NEW)
+	}
+
+	function handleEthClick() {
+		goto(routes.ETH_NEW)
+	}
+
+	function handleImportClick() {
+		console.log('Import from file')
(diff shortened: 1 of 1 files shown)


swarm-ui/src/lib/components/ui/loader.svelte
```diff
@@ -0,0 +1,59 @@
...
+	type Color = 'high' | 'low'
+	type Props = {
+		dimension?: Dimension
+		color?: Color
+	}
+	let { dimension = 'default', color = 'low' }: Props & HTMLAttributes<HTMLDivElement> = $props()
+
+	let size = dimensionToSize(dimension)
+
+	function dimensionToSize(dimension: Dimension): number {
```
vojtechsimetka: This could make it even better and used for icons as they often have specific size variants.
```suggestion
	function dimensionToSize(dimension: Dimension): 32 | 24 | 16 {
```

swarm-ui/src/routes/(app)/connect/+page.svelte
```diff
@@ -61,33 +59,32 @@
...
 
-		// Get parameters from URL
-		const origin = $page.url.searchParams.get('origin')
-		if (!origin) {
-			error = 'No origin parameter found in URL'
-			return
-		}
+		if (!sessionStore.data.appOrigin) {
+			// Get parameters from URL
+			const appOrigin = page.url.searchParams.get('origin')
```
vojtechsimetka: I wonder if we actually need this as a search param and not just hash param since this is single page app. Clearing search parameters causes re-render unlike hash params.

PR #91, opened 2025-12-17
fix: include a number in default account name

### LINKED ISSUE #78: Account default name should include a number
The default name for accounts is `Ethereum` and `Passkey`, based on their types. I suggest that if there is already an account with the default name, then add an increasing number at the end (e.g. the second ethereum account will be `Ethereum 2`).

agazso:
Fix #78


vojtechsimetka (approved):
This does not prevent me to create:
1. Passkey 3
2. Passkey
3. Passkey => Passkey 3 (this function turns it to Passkey 3)

But we can for sure live with this.

Issue #124, opened 2026-01-20
In local development environment the localStorage has different content


agazso:
See the original problem description in #40 

To recap, the issue is that the localStorage is partitioned between the app domain and the swarm-id domain, but in the case of the local development environment (where swarm-id domain is `localhost:<port>`) the content of the localStorage is different when the app has a swarm-id iframe and from when the swarm-id is opened in a browser tab. This is not the case with a proper secure context (e.g. in the deployed version), there the swarm-id domain localStorage is properly shared between the iframe and the browser tab.

Since #123 there is a workaround employed that opens a window of swarm-id for authentication (that has access to the real data in localStorage) and the opened window sends a `setSecret` message to the SwarmIdProxy with all the necessary information required for the client to be able to operate (e.g. postage stamps, app secret etc.). Unfortunately this is not future-proof, if the SwarmIdClient is extended with functionality that requires access to the localStorage then it would need to be added to the `setSecret` message.

It would be better to come up with a solution that emulates access to the `localStorage`. One idea can be to always open a window for the swarm-id domain and keep it open (in the local development environment only) so that it can communicate its content with the app page in a transparent manner.

PR #130, opened 2026-01-21
feat: connect function

agazso:
This PR adds a custom `connect` function to `SwarmIdClient` so that custom connect buttons can be implemented with it, not only the button in the iframe can be used. 

The way it works is that it opens a new window and it does not require the new window to send a `setSecret` message back to the app window, but instead the storage events are listened to and when the storage state changes on the swarm-id page, the proxy iframe on the app window will get a notification.

Please note that due to #124 this does not work in a local development environment, so the old behavior is also kept optionally and can be turned on by passing the `proxyMode`  parameter in the URL. In that case a `setSecret` message is sent. The iframe version uses this by default and that can be used for connecting locally. Unfortunately the custom connect button is not working in the local development environment.

### DIFF
--- a/demo/demo.html
+++ b/demo/demo.html
@@ -46,6 +46,7 @@
         padding: 20px;
         background: #f8f9fa;
         border-radius: 8px;
+        gap: 8px;
       }
 
       .section h2 {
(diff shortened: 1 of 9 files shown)


vojtechsimetka (approved):
Very nice!

Issue #60, opened 2025-12-10
Favicon and link preview


agazso:

david-gauquelin:
is there a way to get the favicon from https://www.ethswarm.org/ ?

david-gauquelin:
Title: Swarm ID - Decentralized Identity
Description: Create and manage your decentralized identity with Ethereum or Passkey. Sync across devices with Swarm network.

Image:

<img width="1200" height="627" alt="Image" src="https://github.com/user-attachments/assets/8355aa7e-25f3-42b6-ba6c-ef671de0f20a" />

agazso:
The problem with the original Swarm favicon is that in dark-mode it's not really visible, because it's black on transparent background. The best practice for using monochrome icons as favicon is to have the alternate color as background (e.g. black on white, or white on black)

PR #157, opened 2026-02-05
fix: favicon now has background

agazso:
Related: #60


agazso:
> But the favicon is now a bit more "zoomed out"

I guess that's by design to have padding/border

PR #172, opened 2026-02-16
feat: feed api

agazso:
This PR adds feed relate API functionality to SwarmIdClient. There are two sets of functions, one for sequential feeds and for epoch type feeds.

It turned out during testing that Bee does not completely support epoch type feeds (see open issues below). More specifically the `/feeds` endpoint only supports sequential feeds. The `/bzz` endpoint supports epoch feed resolution in theory, but I could never get it to work and there seems to bugs in it (according to Claude). 

For now I just removed the ability from the demo app to be able to make Bee compatible epoch feeds, until the bugs are confirmed (and potentially fixed). The sequential feeds work, although in order to make it work with Bee they are using the `/soc` endpoint to upload, not the usual encrypted-data on the `/chunks` endpoint approach.

Also updated the `docs-site`, including the unrelated, missing ACT documentation as well.

Bee issues:
- https://github.com/ethersphere/bee/issues/5028
- https://github.com/ethersphere/bee/issues/4830

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -1478,15 +1478,6 @@ Secret message for ACT test!</textarea
 				return ownerRaw.startsWith('0x') ? ownerRaw.slice(2) : ownerRaw
 			}
 
-			const ensureDownloadIndex = (options) => {
-				const indexInput = document.getElementById('feed-index')
-				const indexRaw = indexInput.value.trim()
-				if (indexRaw) {
-					return options
-				}
-				return options
-			}
-
 			const incrementPayloadValue = (value) => {
 				const match = value.match(/^(.*?)(\d+)\s*$/)
 				if (!match) {
@@ -1861,7 +1852,7 @@ Secret message for ACT test!</textarea
 					const topic = document.getElementById('feed-topic').value.trim()
 					const encryptionKeyInput = document.getElementById('feed-encryption-key')
 					let encryptionKey = encryptionKeyInput.value.trim()
-					const options = ensureDownloadIndex(getFeedDownloadOptions())
+					const options = getFeedDownloadOptions()
 					const owner = getFeedOwner()
 
 					if (downloadBtn.disabled) return
@@ -2072,7 +2063,7 @@ Secret message for ACT test!</textarea
 					const topic = document.getElementById('feed-topic').value.trim()
 					const encryptionKeyInput = document.getElementById('feed-encryption-key')
 					let encryptionKey = encryptionKeyInput.value.trim()
-					const options = ensureDownloadIndex(getFeedDownloadOptions())
+					const options = getFeedDownloadOptions()
 					const owner = getFeedOwner()
 
 					if (downloadBtn.disabled) return


vojtechsimetka (approved):
I'm going to approve this because I did not find anything wrong trying this. Below are things I found relevant from claude review

## Critical

### 2. `writer: undefined as any` in production code

**File:** `lib/src/swarm-id-proxy.ts:1634`, `3290`, `3310`

```typescript
writer: undefined as any, // not used by stamper.stamp
```

Appears 3 times. Violates both **"never use `any`"** and could cause runtime errors if `stamper.stamp()` ever accesses `.writer`. Either provide a real writer or make the field optional in the stamper interface.

---

## Major

### 4. `null` usage violates CLAUDE.md

**File:** `demo/index.html:2152`

```javascript
let lastFeedManifestReference = null
let lastFeedManifestTopic = null
let lastFeedManifestOwner = null
let lastFeedManifestType = null
// also: return null in validateHex()
```

CLAUDE.md says **"Never use `null` — use `undefined`"**.

### 6. Significant code duplication

**File:** `lib/src/proxy/feeds/epochs/async-finder.ts`

- `getEpochChunk` (line ~147) vs `getEpochChunkWithMetadata` share nearly identical SOC parsing logic. `getEpochChunk` should delegate to `getEpochChunkWithMetadata` and extract `.reference`.
- `findPreviousLeaf` (line ~437) is copy-pasted identically in both `async-finder.ts` and `finder.ts`. Should be extracted to a shared utility.

### 8. Inconsistent API parameter ordering

**File:** `lib/src/types.ts:465`

```typescript
// SequentialFeedReader
downloadPayload: (encryptionKey: Uint8Array | string, options?) => ...
downloadRawPayload: (options?, encryptionKey?: Uint8Array | string) => ...
```

`downloadPayload` takes `encryptionKey` as a required first positional arg; `downloadRawPayload` takes it as an optional second arg. Meanwhile, `EpochFeedReader` uses an options object pattern. The APIs should be consistent.

---

## Minor

### 9. `reader.download()` may not exist

**File:** `demo/index.html:2382`

```javascript
const result = await reader.download({ at: BigInt(at) })
```

The `FeedReader` interface in `types.ts` only defines `downloadReference()` and `downloadPayload()`, not `download()`. This would fail at runtime.

### 10. `TimestampSchema` doesn't accept `bigint`

**File:** `lib/src/types.ts:43`

```typescript
export const TimestampSchema = z.union([z.number(), z.string()])
```

But interfaces accept `bigint | number | string`. Since `bigint` can't be serialized via `postMessage`, this could cause Zod validation failures. Worth adding explicit handling or documentation.

### 11. Silent payload length handling

**File:** `lib/src/proxy/feeds/epochs/async-finder.ts:213`

```typescript
const hasSpanPrefix = payload.length === 48 || payload.length === 80
```

Payload format detection assumes exactly 40, 48, 72, or 80 bytes. Unexpected lengths are silently processed as no-span-prefix format, which could produce corrupted data. Should throw or warn.

---

## Nits

- **Docs:** `BigInt(Date.now() / 1000)` should be `BigInt(Math.floor(Date.now() / 1000))` for correctness (`docs-site/src/content/docs/api/index.mdx`)
- **Demo:** Empty `catch {}` blocks silently swallow errors in multiple places
- **Types:** `encryptionKey` validated with `PrivateKeySchema` is semantically misleading; a dedicated `EncryptionKeySchema` alias would be clearer
- **Naming:** Epoch results use `socAddress`, sequential results use `reference` for the same concept — inconsistent

PR #186, opened 2026-02-26
feat: agent account

### LINKED ISSUE #184: Add new account type for agents and automated testing

agazso:
Currently it is hidden by default, but can be enabled from the demo app (and in the `/connect` url).

Closes #184

### DIFF
--- a/swarm-ui/tests/home.test.ts
+++ b/swarm-ui/tests/home.test.ts
@@ -2,5 +2,6 @@ import { test, expect } from '@playwright/test'
 
 test('home page loads successfully', async ({ page }) => {
 	await page.goto('/')
-	await expect(page.locator('h1')).toContainText('Swarm Identity')
+	// When no accounts exist, redirects to account creation page
+	await expect(page.locator('h4')).toContainText('Swarm ID')
 })


swarm-ui/src/lib/components/account-selector.svelte
```diff
@@ -28,7 +29,8 @@
 		accounts.map((account) => ({
 			value: toPrefixedHex(account.id),
 			label: account.name,
-			icon: account.type === 'passkey' ? PasskeyLogo : EthereumLogo,
+			icon:
+				account.type === 'passkey' ? PasskeyLogo : account.type === 'agent' ? Bot : EthereumLogo,
```
vojtechsimetka: I'd prefer for this to not be ternary operator but rather `switch` (wrapped in a function)

swarm-ui/src/lib/components/enter-seed-modal.svelte
```diff
@@ -0,0 +1,144 @@
...
+		if (!seedPhrase.trim()) {
+			error = 'Please enter your seed phrase'
+			return
+		}
+
+		const validation = validateSeedPhrase(seedPhrase)
+		if (!validation.valid) {
+			error = validation.error
+			return
+		}
```
vojtechsimetka: `validateSeedPhrase` could handle the empty string as well

PR #298, opened 2026-04-07
feat: subsidised gateway

### LINKED ISSUE #295: Use SwarmID with gateway postage batch
From this issue: https://github.com/snaha/swarm-id/issues/290#issuecomment-4175747597

> Ideally it should work with gateway-managed stamps also, to the extent that for example feeds could be written using the underlying feed private key of the Swarm ID, but using the Swarm Gateway to actually write to the feed.
This works currently in existing Swarm applications and it makes it easier to do quick onboarding and let users to try out Swarm without blocked by the need of getting their own stamp.

agazso:
This PR adds an optional subsidised gateway URL to be passed to `SwarmIdClient` and let the user upload even if they have no postage stamps.

Closes #295

The rules are not so simple to understand, especially when taking the storage partitioned state into consideration. Therefore there is a new section in documentation updating this feature. Below some excerpts to help understanding when it is possible to upload.

| Storage Partitioned | Has User Stamp | Custom Node | Subsidised Configured | Can Upload | Upload Mode   |
| ------------------- | -------------- | ----------- | --------------------- | ---------- | ------------- |
| Yes                 | —              | —           | Yes                   | Yes        | `subsidised`  |
| Yes                 | —              | —           | No                    | No         | `unavailable` |
| No                  | Yes            | —           | —                     | Yes        | `user-stamp`  |
| No                  | No             | Yes         | —                     | No         | `unavailable` |
| No                  | No             | No          | Yes                   | Yes        | `subsidised`  |
| No                  | No             | No          | No                    | No         | `unavailable` |

### Understanding the Table

**Storage Partitioned (Yes)**

When the browser partitions storage (Safari ITP), the proxy iframe cannot access:

- The user's postage stamps from localStorage
- The custom Bee node URL setting

In this case, only the `subsidisedGatewayUrl` (passed by the dApp at initialization) determines upload capability. This is why subsidised gateways are especially important for Safari support.

**Has User Stamp**

The user has configured a postage stamp in the identity provider. User stamps always take priority over subsidised gateways when available.

**Custom Node**

The user has configured a custom Bee node URL in the identity provider's network settings. When a custom node is set, the subsidised gateway is automatically disabled to ensure uploads go to the user's chosen node.

**Subsidised Configured**

The dApp passed a `subsidisedGatewayUrl` when creating the SwarmIdClient.

## Priority Rules

1. **User's own stamp** — When available and storage is not partitioned, the user's configured stamp is always used
2. **Custom Bee node** — If the user has configured a custom Bee node, the subsidised gateway is disabled (even if configured)
3. **Subsidised gateway** — Used as a fallback when no user stamp is available

This ensures users retain control over their uploads while providing a fallback for those without stamps.


lib/src/proxy/upload-subsidised.ts
```diff
@@ -0,0 +1,341 @@
...
+}
+
+/**
+ * Convert Uint8Array to hex string.
+ */
+function uint8ArrayToHex(bytes: Uint8Array): string {
+  return Array.from(bytes)
+    .map((b) => b.toString(16).padStart(2, "0"))
+    .join("")
+}
```
vojtechsimetka: Duplicate in `utils/key-derivation.ts`

vojtechsimetka (commented):
Good few findings:
- I would consider naming this differently in general because it does not necessarily need to be a gateway. You can have your own nodes with stamps.  
- Feeds don't work in the subsidized mode but SOC and GSOCs do
- Please remove the logs and only keep error logs

agazso:
>     * I would consider naming this differently in general because it does not necessarily need to be a gateway. You can have your own nodes with stamps.

I chose the naming because in order to use this feature you need either a [swarm-gateway](https://github.com/ethersphere/swarm-gateway) or a reverse proxy in front of Bee that adds the `swarm-postage-batch` header to all requests.

Having your own node with stamps is a different feature that we may consider adding later. The biggest issue with that it is not clear what happens if you cannot access the node. For example adding `http://localhost:1633` may work when you are on your computer running a light node locally, but it will break when you are on your phone.

This feature is about making it possible for applications to provide a subsidised network access, but the user preference can override it. Running your own node with stamps would be something that should be specified as user preference per device, potentially with some fallback if you are on a device where your own node is not available. (opened an issue about this: #304)

With that said, we may consider calling this `subsidised node` instead of `subsidised gateway`.

agazso:
The feeds are now fixed and also refactored many repetitive error prone parts of the proxy, such as the write locking, the uploading, message sending etc.

It became a bigger change than I anticipated at first but also removed/fixed a lot of technical debt and the code is in a much better shape for adding functionality in the future.

vojtechsimetka (approved):
Tested and it works. I went through the code and did not find anything obvious but it is a rather large PR.

PR #318, opened 2026-05-13
feat: require authentication before deleting an account

### LINKED ISSUE #241: Ask authentication before deleting an account

vojtechsimetka:
## Summary

- Account deletion now requires fresh authentication before any local data is removed.
- Routes per account type via the existing `getMasterKeyFromAccount` helper:
  - **Passkey** — WebAuthn re-auth
  - **Ethereum** — wallet connect + signature
  - **Agent** — `EnterSeedModal` collects the seed phrase (handled via `SeedPhraseRequiredError`)
- Auth failures keep the confirm modal open with an inline error so the user can retry or cancel; only on success does deletion run (clears identities, apps, stamps, account, session, then navigates home).
- `DeleteModal` gained an optional `error` prop for the inline message; confirm-dialog copy updated to mention the auth step.

Closes #241

## Test plan

- [x] Passkey account: confirm WebAuthn prompt appears on delete; cancel leaves account intact, success deletes and routes home.
- [x] Ethereum account: wallet signature is required; rejecting it shows the inline error and keeps the modal open.
- [x] Agent account: confirm `EnterSeedModal` opens; invalid seed re-opens the delete modal with an error; valid seed deletes the account.
- [x] Re-opening the delete modal after a prior error starts clean (no stale error text).

### DIFF
--- a/swarm-ui/src/lib/components/delete-modal.svelte
+++ b/swarm-ui/src/lib/components/delete-modal.svelte
@@ -50,8 +50,12 @@
       <Typography style="color: var(--colors-red)">{error}</Typography>
     {/if}
     <section class="buttons">
-      <LoaderButton variant="strong" dimension="compact" onclick={confirm} class="danger-button"
-        >{buttonTitle}</LoaderButton
+      <LoaderButton
+        variant="strong"
+        dimension="compact"
+        onclick={confirm}
+        stayActive={false}
+        class="danger-button">{buttonTitle}</LoaderButton
       >
       <Button variant="ghost" dimension="compact" onclick={oncancel} disabled={disableCancel}
         >{cancelTitle}</Button
--- a/swarm-ui/src/lib/components/delete-modal.svelte
+++ b/swarm-ui/src/lib/components/delete-modal.svelte
@@ -16,6 +16,7 @@
     text: string
     buttonTitle: string
     cancelTitle?: string
+    error?: string
   }
   let {
     oncancel,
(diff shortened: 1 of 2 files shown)


agazso:
There is a potential race condition when canceling. User clicks "Delete account" → external auth prompt (passkey biometric / wallet signature) appears. While that's pending, the modal's "Don't delete" button is still clickable (LoaderButton only covers the confirm button, not Cancel). Clicking Cancel closes the modal but does not abort the in-flight handleDeleteAccount. If auth then completes successfully (cached passkey assertion, pre-approved wallet, reflexive fingerprint touch), performAccountDeletion() runs and the account is deleted anyway. This is the exact failure mode the gate is meant to prevent. Fix: track a cancel flag and check before performAccountDeletion, or disable Cancel while LoaderButton is loading. Can be done as a separate issue/PR.

agazso (approved):

vojtechsimetka:
Good catch — fixed in 48809229.

- Track `isDeleteAuthInProgress` reactive state and a non-reactive `deleteCancelled` flag.
- All cancel paths (Cancel button, close icon, Esc, click-outside) flow through the modal's `oncancel`; if auth is still pending, they set `deleteCancelled = true`.
- After `getMasterKeyFromAccount` resolves we check the flag and bail before `performAccountDeletion`, so a late-resolving passkey/wallet promise can't delete the account.
- Also pass `disableCancel={isDeleteAuthInProgress}` to disable the Cancel and Close buttons while auth is in flight, making the busy state explicit. (Esc/click-outside still work but are caught by the flag as a backstop.)

PR #325, opened 2026-05-27
feat(stamps): compute dev-page stamp amount from chainstate

### LINKED ISSUE #326: Dev page "Buy Postage Stamp" fails on bee-compose with 400 "insufficient validity"
Clicking **Buy Stamp** on `/dev` against a bee-compose Bee node returns:

> `{"code":400,"message":"insufficient amount for 24h minimum validity"}`

with Bee logs reading:

> `insufficient initial balance for 24h minimum validity. balance 100000000, minimum amount: 414720000`

The dev page hardcoded the `amount` parameter (most recently `500_000_000` PLUR/chunk) and posted it to Bee's `POST /stamps/{amount}/{depth}` without ever asking the chain what the price actually is.

Bee enforces `amount ≥ currentPrice × blocksPerDay` (17,280 blocks at 5 s/block) so the deposit covers at least 24 h. On `bee-compose` the PriceOracle sits at Bee's hardcoded floor of `24_000` PLUR/chunk/block = Bee's contract default for a dev chain with no traffic. The 24 h minimum therefore is:

`24_000 × 17_280 = 414_720_000` PLUR/chunk

This is not a `bee-compose` bug — mainnet today is ~`57_100` PLUR/chunk/block (≈2.4× the floor), so the same hardcoded default would also break against mainnet. `fdp-play` happened to be configured below the standard floor, which is why this only surfaced after the cluster switch.

vojtechsimetka:
## Summary

- Replace the hardcoded \`500_000_000\` default amount on the dev page with one derived live from Bee's \`GET /chainstate\`. The old default silently failed against bee-compose (PriceOracle floor of \`24_000\` PLUR/chunk/block requires \`≥ 414_720_000\` PLUR/chunk for 24h) and would break again on any chain with a different price config.
- Add \`fetchChainState\` and \`calculateStampAmountForDays\` to \`@snaha/swarm-id\` so callers compute the right amount from the chain instead of hand-bumping a magic number.
- Surface the current chain price and Bee's 24h minimum in the dev UI so the constraint is visible, not a cryptic 400.

## Why

User hit \`insufficient initial balance for 24h minimum validity. balance 100000000, minimum amount: 414720000\` on bee-compose after switching from fdp-play. Root cause is that the dev page never asks Bee what the chain price actually is — it sends a baked-in number. The fix is to read \`currentPrice\` from \`/chainstate\` and size the deposit so it clears Bee's enforced 24h floor.

## What's in the lib

- \`fetchChainState(beeUrl)\` — parses \`block\` + \`currentPrice\` (BigInt, since Gnosis values can exceed \`Number.MAX_SAFE_INTEGER\`). Throws on missing fields so callers can surface the error.
- \`calculateStampAmountForDays(currentPrice, days)\` — \`currentPrice * BLOCKS_PER_DAY * days\`. \`days = 1\` returns exactly Bee's reported minimum.
- \`BLOCKS_PER_DAY\` constant (\`17280n\`) and \`ChainState\` type exported.
- 10 new unit tests, including one pinning the bee-compose minimum (\`24_000n × 17_280n = 414_720_000n\`) against Bee's actual error message.

## What's in the dev page

- Amount input starts empty; on first chainstate load, autofills with 7 days' worth at the current price. Subsequent Bee URL changes refresh the displayed price but don't clobber user edits.
- Helper line under Amount: \`Chain price: X PLUR/chunk/block · 24h min: Y PLUR · default fills 7d validity\`, or a red error if \`/chainstate\` fails (e.g. Bee unreachable).
- Buy button is disabled when the amount field is empty.

## Test plan

- [x] \`pnpm check:all\` clean (485 lib tests pass; svelte-check, eslint, prettier, knip all green)
- [ ] Reviewer: with bee-compose running, open \`/dev\` → Stamps tab → confirm Amount autofills to \`2,903,040,000\` and the buy succeeds
- [ ] Reviewer: kill bee node, switch Bee URL → confirm helper text flips to red error message
- [ ] Reviewer: edit Amount manually, then change Bee URL → confirm the manual value is preserved (chain price line updates, input does not)

### DIFF
--- a/lib/src/utils/ttl.ts
+++ b/lib/src/utils/ttl.ts
@@ -11,6 +11,13 @@
 export const GNOSIS_BLOCK_TIME = 5
 
 /**
+ * Blocks per day on Gnosis Chain (24h / 5s block time = 17280). BigInt so it
+ * composes with the PLUR-denominated price math in
+ * {@link calculateStampAmountForDays} without precision loss.
+ */
+export const BLOCKS_PER_DAY = (24n * 60n * 60n) / BigInt(GNOSIS_BLOCK_TIME)
+
+/**
  * Time constants
  */
 const SECONDS_PER_MINUTE = 60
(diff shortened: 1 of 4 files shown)


swarm-ui/src/routes/(app)/dev/+page.svelte
```diff
@@ -313,9 +326,27 @@ Check console logs for details:
...
+  async function loadChainState() {
+    chainStateError = ''
+    try {
+      const state = await fetchChainState(beeUrl)
+      currentPrice = state.currentPrice
+      // Only autofill if the user hasn't typed (or pre-typed) a value yet.
+      // After the first fill, the user owns this field — switching bee URLs
+      // updates the displayed price but does not clobber their input.
+      if (stampAmount === '') {
+        stampAmount = calculateStampAmountForDays(state.currentPrice, DEFAULT_STAMP_DAYS).toString()
```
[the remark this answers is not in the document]
vojtechsimetka: This is dev only page, no need to fix

lib/src/utils/ttl.ts
```diff
@@ -130,3 +137,77 @@ export function calculateExpiryTimestamp(
...
+ * whose amount does not cover at least 24h of storage at `currentPrice`.
+ *
+ * @param beeUrl - Bee node URL
+ * @returns Parsed chain state with `currentPrice` as BigInt
+ * @throws if the request fails, returns non-OK, or omits required fields.
+ *         Callers (e.g. the dev page) surface this directly to the user.
+ */
+export async function fetchChainState(beeUrl: string): Promise<ChainState> {
+  const base = beeUrl.replace(/\/$/, "")
+  const response = await fetch(`${base}/chainstate`)
```
agazso, line 166: This will not work on gateways

agazso (approved):
There are some minor issues found by Claude, more like UX improvements, nothing is a blocker.

 - Auto-fill only when empty — `if (stampAmount === '')` is the right gate. Subsequent URL changes update the price display without clobbering user input.
  - No loading indicator — While `currentPrice === undefined && chainStateError === ''` (the in-flight window), the helper text section is invisible. On a slow connection a user might click "Buy" on an empty field before the fill arrives. Since the button is disabled on !stampAmount this is safe, just slightly confusing. A "Fetching chain price…" placeholder would help UX; not blocking for a dev page.
  - No retry path — If /chainstate fails (red error shown), the only way to retry is to change the Bee URL. A "Retry" link next to the error would be   cleaner.
  - `BigInt.toLocaleString()` — Works in all modern browsers and Node 18+, but if this component ever runs in an older environment it will throw. Fine for a dev page.
  - Effect coupling — `loadChainState()` is always called alongside `loadBeeStamps()` in the same $effect branch. This means chain state is never refreshed independently of stamps. Acceptable for a dev page but worth noting if this page grows.
  - `calculateStampAmountForDays(0n, 7)` returns `0n` — If Bee ever reports currentPrice: "0", stampAmount is set to "0" (truthy), the button is enabled,
  and the buy will fail at Bee. This is a genuinely impossible real-world case on Gnosis, but worth guarding in the auto-fill:
```javascript
  const amount = calculateStampAmountForDays(state.currentPrice, DEFAULT_STAMP_DAYS)
  if (stampAmount === '' && amount > 0n) {
    stampAmount = amount.toString()
  }
```

lib/src/utils/ttl.ts
vojtechsimetka: Good catch — leaving as-is though: on a gateway `fetchChainState` just throws, the catch in `loadChainState` surfaces it as the red "Could not fetch chainstate from Bee" helper, and `stampAmount` stays empty. User can still type an amount manually; the Buy button re-enables as soon as the field is non-empty. And gateways wouldn't accept `POST /stamps` anyway, so degrading to manual entry is the right end state.

Issue #217, opened 2026-03-13
Support multiple devices with synced account at the same time


agazso:
# Research: Multi-Device Postage Stamp Coordination

## Problem Statement

The `UtilizationAwareStamper` in `lib/src/utils/batch-utilization.ts` tracks 65,536 buckets (Uint32Array), each with up to `2^(batchDepth-16)` slots (e.g., 16 at depth 20, 256 at depth 24). When multiple devices/tabs share the same identity and postage batch, each maintains its own local copy of these counters (cached in IndexedDB). Without coordination, two devices can assign the same bucket:slot pair, causing:

1. **Data loss** — mutable stamps allow overwriting, so the second write silently replaces the first
2. **Incorrect utilization state** — each device's counter diverges from reality

### Key Constraint: Dynamic Active Device Count

A user may have many registered devices (laptop, phone, desktop, iPad, etc.) but typically only **~2 devices write simultaneously**. This means:

- **Static interleaving by total registered devices is too wasteful** — if 6 devices are registered but only 2 are writing, interleaving by 6 wastes ~67% of slot capacity
- At batch depth 20 (16 slots/bucket), interleaving by 6 leaves only ~2 slots per device per bucket
- **Real-time presence detection is needed** to know which devices are _actively writing_, so interleaving can be dynamic (e.g., interleave by 2 when 2 are active)

vojtechsimetka:
A code review of the sync coordination surfaced three concrete correctness bugs under this umbrella:
- #413 — `PartitionLease.refresh()` re-mints the lock generation every tick, which inverts claim order under a real (advancing) clock and can make both holders demote. Masked today because every lease test freezes the clock.
- #414 — the state pointer is published in the same `Promise.all` as the chunks it references, so a partial PUT failure + crash can leave a dangling pointer that bricks the partition read-only until batch expiry.
- #415 — feed finders / roster append treat read failures as a free slot, enabling peer-entry overwrite on a flaky gateway.
- #427 — the test gaps that let #413/#414/#415 through (advancing-clock dual-hold, crash-mid-publish, concurrent roster-append).

PR #335, opened 2026-06-04
feat: multi-device postage batch sharing + account-state sync

agazso:
## Multi-device postage batch sharing + account-state sync

Lets several devices that share one Swarm ID account write to a **single mutable postage batch** concurrently — and keeps the shared account snapshot consistent across them — using only Swarm primitives (chunks, SOCs, epoch feeds). No server, no coordinator.

### What's in this PR

Two independent mechanisms:

- **Partition lease** — each active device leases a *partition* of every bucket's slot space (`slot = K + partition + K·j`, `K = PARTITION_COUNT = 2`) so two devices never stamp the same `(bucket, slot)`. A per-partition **lock SOC** (owned by the account's backup signer, generation-fenced) is the cross-device authority for "who holds this partition"; a per-partition **state feed** carries the per-bucket counter so a taking-over device resumes from exactly where the previous holder left off. Lease lifecycle covers first device, second device, 3+ device turn-taking, and crash recovery.
- **Account-state sync** — one epoch feed per account (`swarm-id-backup-v1:account:<id>`) holding an encrypted snapshot (identities / connected apps / stamps / devices). Publishing uses an **optimistic verify-retry** (re-read the feed, re-merge, republish up to 3× with jittered backoff) to survive concurrent writers on a feed with no compare-and-swap, plus a **full-snapshot refresh** path and **cross-context publish** so a connection made in a transient popup is published by a persistent window. **Connected-app revocation** now propagates via last-writer-wins + tombstones (`updatedAt` / `revokedAt`).

Also: `/dev` now buys **mutable** batches (the partition lock SOC must be re-writable), and the stamper derives its depth from the stamp instead of a hardcoded default.

### How it works / porting

The full implementation-level spec — slot layout, lock-SOC protocol, feeds, key derivation, lease lifecycle, merge rules, and a porting checklist for re-implementing the scheme on another platform (native mobile, desktop) — lives in the new docs-site page:

- **[Multi-Device Postage Batches](https://github.com/snaha/swarm-id/tree/feat/postage-batch-partitioning/docs-site/src/content/docs/multi-device-postage-batches.mdx)** (renders under docs → Multi-Device → Postage Batch Sharing)

This PR also consolidates the scattered `docs/Multi-Device-*` working notes into that single page. `docs/BatchWriteCoordinator-Design.md` is kept as the design for the follow-up refactor.

### Tested

- Unit tests (lib): `pnpm --filter @snaha/swarm-id check:all` — 647 tests green, incl. partition-lock / lease integration and connectedApps merge (LWW + tombstones).
- `pnpm --filter swarm-identity check:all` — lint / svelte-check / knip green (0 errors).
- `pnpm build:docs` — new page builds and renders.
- Manual multi-device E2E against a local bee cluster (Playwright diagnostic `swarm-ui/tests/multi-device-connect-sync.test.ts`, kept as a local-only diagnostic — it hardcodes a local batch/signer/seed and needs the dev cluster).

### Next steps

All tracked under the multi-device umbrella (#217):

- **`BatchWriteCoordinator` refactor + proxy-side publish** (#336) — extract the duplicated partition-lease / stamp / write-lock orchestration into one shared unit, then publish account state from the proxy (incl. announcing a device on lease acquisition); also folds in clearer contention logging.
- **Restore / refresh coverage** (#338) — agent-account restore-from-Swarm + broaden refresh triggers to app load / account switch.
- **Merge consolidation + deletion tombstones** (#337) — one shared snapshot-merge implementation, and extend the connectedApps tombstone pattern to identities & stamps.


vojtechsimetka (approved):
Tested it locally and it works fine as far as I can tell — multi-device batch sharing and account-state sync behave as expected for me.

Approving. Below are things that could change code-wise (mostly LLM-flagged, none blocking — migration/history/data-loss is explicitly out of scope since we have no users yet).

### Duplication
- **Merge logic exists in two places.** `lib/src/sync/merge-snapshot.ts` is re-implemented almost verbatim in `swarm-ui/src/lib/utils/refresh-account-from-swarm.ts:39-73` (`unionByKey` / `mergeConnectedAppsLww` / `mergeDeviceLists`) — the comments literally say "mirrors … merge-snapshot.ts". The refresh path can build a local snapshot and call `mergeSnapshotWithRemote`, then apply. Two copies of the subtle LWW-tombstone rule will drift. (Tracked in #337 — worth pulling forward, it's cheap.)
- **Partition orchestration duplicated.** `lib/src/sync/sync-account.ts:400-477` hand-rolls "read all locks → pick held/free → claim", which `PartitionLease` already does — and its inline claim seeds no `self`/synced-reference, so "held" means something different there than in the proxy. A read-only `PartitionLease` would unify it. (Tracked in #336 — the thing to extract is the *semantics*, not just the LOC.)

### Dead code (free deletions)
- `PartitionStateSchemaV1` / `PartitionState` in `lib/src/schemas.ts:311` — unused, **and** it describes a wire format that doesn't match what `partition-state.ts` actually writes (reference-chunk → counter-chunks). Misleading; remove it.
- `PartitionLease.partitionFor()` — identical to `heldPartition()`, no caller.
- `resolveUtilizationChunkKeys` — exported, no production caller (only `chooseUtilizationChunkKey` is used).

### Code-quality nits
- `acquirePartitionLock` (`lib/src/sync/partition-lock.ts:238`) returns `verified` on the acquired branch even when it's a foreign payload, so `self` lease timestamps end up sourced from a peer. Return `ourPayload` when `verified.holderDeviceId !== deviceId`.
- The `lib/src/proxy/upload.ts` Merkle-span fix touches **all** uploads and I don't see a roundtrip test forcing ≥2 tree levels — worth adding one (plain + encrypted).
- `calculateUtilization` spreads 65,536 args into `Math.max(...)` — swap for a loop.
- Tooling changes are bundled in (`package.json`: pnpm 10→11, dropped `onlyBuiltDependencies`) — unrelated to the feature, ideally its own PR.

🤖 Findings mostly generated with Claude Code.

agazso:
Follow-up to the partition-lease schema work: the binary **partition-state** wire format still needs a schema + decode-time validation — tracked in #340.

The partition-**lock** SOC payload is done in 6c582757 — `PartitionLockPayloadSchemaV1` + read-side `safeParse` validation (a foreign/corrupt/old payload is now treated as "lock unobserved" instead of an unchecked cast). Versioned by the existing `swarm-id-partition-lock-v1` identifier domain, so no wire-format change.

PR #346, opened 2026-06-09
feat: extract a BatchWriteCoordinator from the proxy

### LINKED ISSUE #336: Refactor: extract a BatchWriteCoordinator from the proxy
Extract the write/lease/lock/stamp concern out of the ~4,350-line `lib/src/swarm-id-proxy.ts` into a reusable `BatchWriteCoordinator`, so the proxy and `sync-account` share one implementation. Full design: [`docs/BatchWriteCoordinator-Design.md`](../blob/main/docs/BatchWriteCoordinator-Design.md).

Follows #335 (multi-device postage batch sharing + account-state sync). Part of #217.

## Motivation

`swarm-id-proxy.ts` mixes three unrelated concerns: (1) client communication (the postMessage protocol — its actual job), (2) auth/identity/storage/button UI, and (3) the write path — stamp management, the cross-tab Web Lock, the partition-lease lifecycle, upload guards. Concern 3 is cohesive and self-contained but trapped inside the proxy, so it can't be reused.

> **Note:** the original design doc motivates this with "`sync-account` takes no write lock and can't acquire a partition, so a SwarmUI change is skipped." That part is now **stale** — `sync-account.ts` already wraps publishing in `withBatchWriteLock` and claims a free/expired partition in its gate (`acquirePartitionLock`), so the partition-gate publish path works. The remaining motivation is:

1. **Shrink the proxy** — lift concern 3 out so the proxy is just communication + auth + UI.
2. **De-duplicate orchestration** — the proxy and `sync-account` now carry two parallel lease/lock/claim implementations; a single shared coordinator removes the drift risk.
3. **Enable proxy-side publish** — the real open gap (see #335 "next steps"). Today an open identity-UI window does the publishing. In production the user is on the dApp, not the identity UI, so the only guaranteed-persistent context is the **proxy iframe** — which already holds a coordinator-like write path but does not publish account state. A shared coordinator lets the proxy publish on a connection change.
4. **Announce a device on lease acquisition.** When the proxy first acquires its partition lease, publish account state so the device's own `metadata.devices` entry reaches peers. Today a device that signs in / becomes active never publishes its own registration unless some *other* SwarmUI mutation triggers a sync (`restoreAccountToStores` → `accountsStore.addAccount` doesn't trigger a sync, and the restore-time `addIdentity`'s `triggerSync` is gated on an unset `currentIdentityId`). **Observed bug:** with 3 devices the newcomer shows all 3, but the other two only ever see 2, because the newcomer never publishes itself. Lease acquisition is the natural trigger — the device provably holds a partition then, so the publish isn't blocked by the partition gate. (Publishing on plain sign-in/restore was considered but rejected: it races the gate and announces even never-active devices.)

## Proposed boundary — `BatchWriteCoordinator`

A class that owns everything needed to **write to a shared postage batch as a partition holder**, and nothing about client communication.

**Moves out of the proxy (concern 3):**
- **Stamp management** — `initializeStamper`, `getOrCreateWorkerPool`, `saveStamperState(IfNeeded)`, `getUploadTarget`; fields `stamper`, `stampWorkerPool`, `stamperAccountFingerprint`, `utilizationStore`.
- **Locking** — `withWriteLock` (the `swarm-write-<batchIdHex>` Web Lock) and `withModeAwareWriteLock`.
- **Lease lifecycle** — `acquirePartitionLease(WithSlotWait)`, `ensurePartitionLease`, `ensureLeaseStillValid`, `isDisplaced`, `scheduleLeaseRefresh`/`refreshTick`, `yieldIdleLease`, `demoteSilently`, `pauseLeaseBackgroundWork`, `tearDownPartitionLease`, `capture/read/writeLeaseCache`; the `partitionLease`, `isReadOnly`, `leaseContext`, timer, and counter fields.
- **Upload guards** — the batch/partition parts of `ensureCanUpload`, `ensurePartitionHeldForUpload`, `isSubsidisedModeActive`.

**Stays in the proxy:** the message protocol, auth/storage/identity, button UI, and the thin `handleX` handlers (which now delegate writes to the coordinator).

## Interface (sketch)

```ts
interface BatchWriteCoordinatorOpts {
  bee: Bee
  batchId: BatchId
  batchDepth: number
  signerKey: string
  account: { owner: EthAddress; encryptionKey: Uint8Array; accountId: string; partitionCount: number }
  deviceId: string
  utilizationStore: UtilizationStoreDB
  subsidisedGatewayUrl?: string
  onLeaseChange?: (partition: number | undefined) => void
}

class BatchWriteCoordinator {
  static async create(opts: BatchWriteCoordinatorOpts): Promise<BatchWriteCoordinator>

  // Single write entry point: take the Web Lock, ensure a held partition
  // (acquire / slot-wait), bind the stamper, run `op`, flush stamper state.
  withWrite<T>(op: (target: UploadTarget) => Promise<T>,
              opts?: { useWorkers?: boolean; workerCount?: number }): Promise<T>

  startLease(): void   // eager acquire + refresh timer (long-lived holder = the proxy)
  release(): Promise<void>
  teardown(): void

  get currentPartition(): number | undefined
  get isReadOnly(): boolean
  get stamper(): UtilizationAwareStamper | undefined  // for appKey / uploadMode
}
```

`onLeaseChange` replaces the proxy's inline `emitConnectionInfoIfChanged()` calls in the lease code: the coordinator fires it on every transition (acquire / adopt / read-only / yield / demote) and the consumer decides what to do (proxy → push `connectionInfoChanged`; identity UI → update its store).

## Consumers

- **Proxy** — construct one coordinator after auth + stamp context is known, then `startLease()`. Upload handlers become `coordinator.withWrite(target => uploadData(target, …))`. `buildConnectionInfo` reads `currentPartition`/`isReadOnly`/`stamper`. Disconnect → `teardown()`.
- **`sync-account`** — construct a coordinator for the account's default batch and publish via `coordinator.withWrite(...)` — **same Web Lock + same acquire**. One-off mode (no `startLease`/refresh timer): `withWrite` acquires a free/expired partition under the lock if none is held; the lease then lapses by TTL.
- **Identity-UI devices page** — unchanged; still a read-only `PartitionLease` for display.

## Shared lock + lease

- The `swarm-write-<batchIdHex>` Web Lock name convention lives in **one** place (the coordinator, or the tiny `withBatchWriteLock` helper it wraps), with a no-`navigator.locks` fallback for node/tests. Every proxy tab and the identity UI serialize on it.
- The proxy and identity UI share one `deviceId` (`getOrCreateDeviceId`), so a lock SOC written by either is recognised as self-held by the other; they cooperate on the same partition rather than contend.

## Counter coherence (main correctness risk)

Two coordinator instances (proxy iframe + identity-UI tab) share the IndexedDB utilisation cache and the `swarm-id-utilization` BroadcastChannel. Correctness relies on each reading the latest counter from the cache and flushing under the Web Lock. The coordinator should make **reload-before-write / flush-after-write under the lock** explicit. The dominant case — bootstrap (first device, no concurrent writer) — has no contention.

## Migration phases (each keeps `pnpm check:all` green)

1. `withBatchWriteLock(batchIdHex, op)` shared helper; the proxy's `withWriteLock` delegates to it. (Largely done — `lib/src/utils/batch-write-lock.ts` exists and `sync-account` uses it; verify the proxy delegates too.)
2. Extract a `PartitionLeaseManager` (the lease cluster + cache + timer); the proxy delegates; behaviour-preserving.
3. Extract stamp management (stamper + worker pool + `getUploadTarget` + `saveStamperState`).
4. Compose into `BatchWriteCoordinator` with `withWrite` + `onLeaseChange`; proxy handlers call it.
5. Wire `sync-account` to the coordinator and **enable proxy-side publish** (publish account state from the proxy **on lease acquisition / a connection change** when no identity-UI window is open — see motivation #4 for the device-announce case). _(The original phase-5 "SwarmUI publish" fix is already shipped in #335; this phase is now about routing both writers through the coordinator and adding the proxy publish.)_

## Incidental cleanups

- **Clearer contention logging (do alongside phase 5).** `sync-account.ts` logs `[SyncCoordinator] Skipping sync …: all partitions are held by other devices` even when the real cause is a stamp/SOC error, not genuine contention. When the partition gate moves into the coordinator, distinguish "every partition is held by a live foreign device" from stamp/SOC failures and log them separately.

## Risks / open questions
- **Lease mutations run outside the write lock (async refresh tick).** `refreshTick` (a bare `setInterval`, not under the write lock) can call `demoteSilently` / `yieldIdleLease` while an upload is mid-stamp. `demoteSilently` calls `invalidateLease()` then `unbindPartition()` — and `unbindPartition()` resets `leaseStale = false`, so `stamp()`'s `partition !== undefined && leaseStale` circuit breaker never fires for the in-flight upload; it silently falls back to legacy slot-picking instead of aborting with `PartitionLeaseLostError`. The breaker only reliably fires on the upload-entry path (`ensureLeaseStillValid`, which runs under the lock). The coordinator should make lease mutations lock-synchronized (or invalidate-without-unbind so in-flight stamps abort), and a multi-device test should exercise displacement-during-upload. (Found in the #335 review.)

- Handlers are tightly coupled to the inline lock+lease; the coordinator must expose enough state for `buildConnectionInfo` (partition, read-only, appKey/uploadMode via the stamper).
- Long-lived (proxy) vs one-off (`sync-account`) lease modes — keep the refresh timer optional.
- Cross-tab counter coherence (above) — the lock + cache-reload/flush discipline must be explicit.
- Large mechanical surface — do it phased, not big-bang.

## Files (when implemented)

- New: `lib/src/sync/batch-write-coordinator.ts` (+ optionally `partition-lease-manager.ts`).
- `lib/src/swarm-id-proxy.ts` — delegate write/lease/lock/stamp to the coordinator; keep communication / auth / UI.
- `lib/src/sync/sync-account.ts` — publish via the coordinator.

agazso:
Closes #336.

Extracts the proxy's write/lease/lock/stamp concern into a reusable `BatchWriteCoordinator` so the proxy and `sync-account` share one implementation, and uses that seam to fix three gaps the issue calls out. Design + rationale: `docs/BatchWriteCoordinator-Implementation-Plan.md`.

## What changed

- **`lib/src/sync/batch-write-coordinator.ts` (new)** owns the cross-tab Web Lock, the partition-lease lifecycle (acquire / slot-wait / refresh / yield / demote), and the stamp flush. Single entry point `withWrite(op, { wait })` serves both consumers: the proxy (`block` — slot-wait then throw) and `sync-account` (`skip` — claim-once then throw `PartitionContendedError`).
- **`swarm-id-proxy.ts` shrank ~530 lines net.** It constructs a persistent coordinator in `initializeStamper`, routes all ~14 upload handlers through it via the thin `withModeAwareWriteLock` adapter (handlers untouched; subsidised mode stays in the proxy), and reads `coordinator.currentPartition` in `buildConnectionInfo`.
- **`lib/src/sync/publish-account-state.ts` (new)** holds the shared publish/merge/verify core, extracted from `sync-account`'s `runSync`.
- **`sync-account.ts`** drops its inline partition gate + outer `withBatchWriteLock` and publishes through a one-shot coordinator. The "skipping … all partitions held" warn is now distinct from operational errors (which surface as `status: "error"`).

## Fixes

1. **Shrink the proxy + de-duplicate** the two parallel lease/claim implementations into one coordinator.
2. **Displacement-during-upload race** (found in the #335 review): `refreshTick` ran off-lock and `demoteSilently` called `invalidateLease()` then `unbindPartition()`, the latter resetting `leaseStale=false` so the `stamp()` circuit breaker never fired — the in-flight upload silently fell back to legacy slot-picking and corrupted the displacing peer's slots. Now split into an immediate `invalidateLease()` (in-flight `stamp()` aborts with `PartitionLeaseLostError`) followed by an unbind under the write lock.
3. **Proxy-side publish / device-announce** (motivation #3 + #4): the proxy now publishes the account snapshot on lease acquisition (and on account-state changes while a lease is held), debounced and deferred outside the acquiring lock. This closes the observed 3-device bug where a newcomer never announced itself, so peers stayed at 2 devices.

## Review-fix hardening (32384e8f)

A critical self-review of this PR surfaced two real gaps plus smaller hardening, all landed in one follow-up commit:

- **Error classification**: `acquire()` no longer swallows operational failures — the skip path throws `PartitionContendedError` only when the acquire *completed* read-only, so a Bee failure during the lock-SOC scan surfaces from `sync-account` as `status: "error"` instead of a quiet "all partitions held" skip.
- **Idle-yield race**: the voluntary yield now runs **under the write lock** with re-checks (disposed / lease identity / `activeUploadCount` / idle window) — previously it could release the partition and unbind the stamper underneath an upload that entered `withWrite` mid-yield (the same corruption class as the displacement race above).
- **Lease-epoch guard**: an acquire that loses the timeout race discards its late result instead of resurrecting a cleared lease and arming a detached refresh timer.
- **Type safety**: `sync-account` narrows the store's stamper with a runtime `instanceof UtilizationAwareStamper` check (clear `status: "error"`) instead of a double-cast.
- **Account-mismatch guard**: the proxy publish bails if the assembled snapshot's account no longer matches the coordinator's, so one account's feed can never be published through another's batch/lease.
- **Announce-once gating + contribution-aware verify**: a reload no longer republishes when the device is already in the feed, and a co-writer whose union still contains our entities is not counted as a feed-slot loss (no verify-retry storm).

## Tests

- `batch-write-coordinator.test.ts` — `withWrite` block/skip fork, `PartitionContendedError`, lease-change/acquired callbacks, the **displacement-during-upload ordering** (invalidate-before-unbind), teardown/disposed safety, operational-error classification (skip and block), idle-yield under the lock (incl. both abort paths), and the acquire-epoch guard.
- `sync-account.test.ts` — reworked to mock the coordinator (claim/skip now lives in the coordinator); covers publish-on-hold, skip-on-contention, and error-on-operational-failure.
- `publish-account-state.test.ts` — device-announce union (a newcomer announces itself without dropping the peer).

Verification: `pnpm check:all` passes (the swarm-ui knip config hints are pre-existing and non-fatal), **680 lib tests pass**.

Manual multi-device verification against a local cluster: **done** — 3 profiles on one account converge to 3 devices, a forced displacement aborts the upload cleanly, and a UI sync re-claiming the proxy's self-held partition causes no lease flapping.


docs/BatchWriteCoordinator-Implementation-Plan.md
vojtechsimetka, line 1: Lets change this into a doc that explains the rationale and implementation but lets not have it as implementation plan with phases.

lib/src/sync/batch-write-coordinator.ts
```diff
@@ -0,0 +1,756 @@
...
+      )
+      this.partitionLease = undefined
+      this.readOnly = false
+      // Keep cache and in-memory state in lockstep: no lease → no live cache.
+      this.deps.writeLeaseCache?.(undefined)
+      // Operational failure (the lock-SOC scan/claim threw) — NOT contention.
+      // Propagate so the skip path reports an error instead of "all partitions
+      // held", and the block path pauses background work on the real cause.
+      throw error
+    }
```
vojtechsimetka, line 474: **`acquire()`'s catch path is missing the lease-epoch guard.**

The success path discards a late-completing acquire (`if (this.leaseEpoch !== epoch) return`), but this catch mutates shared state unconditionally. Sequence:

1. The block-path acquire races the 45s timeout; the timeout wins → `pauseLeaseBackgroundWork()` bumps the epoch, and the original `acquireWithSlotWait()` continues detached (`Promise.race` can't cancel it).
2. A later `withWrite` re-acquires successfully — new `partitionLease`, cache written, refresh timer armed.
3. The detached acquire's `lease.acquire()` finally **rejects** → this catch clears `this.partitionLease`, resets `readOnly`, and wipes the lease cache — clobbering the *new* live lease's in-memory state while its refresh timer stays armed.

It self-heals on the next write (re-acquire), but it's the same resurrect/clobber class the epoch guard was added to prevent — just mirrored to the failure side. Suggest guarding the catch too:

```ts
} catch (error) {
  // A newer lease owns the state now — don't clobber it on a stale failure.
  if (this.leaseEpoch !== epoch) throw error
  this.partitionLease = undefined
  ...
```

The acquire-epoch-guard test covers the late-*resolve* path only; a late-*reject* sibling test would pin this down.

vojtechsimetka (approved):

lib/src/sync/batch-write-coordinator.ts
agazso: Good catch — confirmed: the catch could clobber a newer live lease's state after the `Promise.race` timeout detached the original acquire. Fixed in 03e91796 by mirroring the epoch guard to the failure side (`if (this.leaseEpoch !== epoch) throw error` before any state mutation), plus the late-reject sibling of the acquire-epoch-guard test — it reproduces exactly this sequence and fails without the guard.

agazso:
Follow-up to the r3395673378 fix: I swept the coordinator for the same stale-continuation class and hardened four more paths in b2aba4a9 (each pinned by a test that fails without the fix):

1. **`refreshTick` staleness guards** — clearing the interval can't stop a tick already in flight (and slow ticks can overlap). A stale tick could re-claim a released partition via `lease.refresh()` (overwriting the release sentinel after teardown), demote a *re-acquired* live lease via the unguarded locked `finalizeDemote` (the exact clobber class of the original finding, on the displacement branch), and resurrect a cleared lease cache in its tail (`writeLeaseCache(lease.serialize())`) that a later acquire would `adoptIfLive` with no Swarm round-trip. Guards now mirror the idle-yield branch: at entry, before arming the breaker, inside the locked demote, and before the tail.
2. **`acquire()` failure unwind** — a consumer callback (`writeLeaseCache`/`onLeaseAcquired`) throwing after `bindPartition`/`startRefreshTimer` left the stamper bound and the interval leaked while the catch dropped the lease record. The catch now clears the timer and invalidates+unbinds.
3. **`acquireWithSlotWait` epoch exit** — a loop detached by the 45 s race timeout would wake from its poll sleep and claim the lock SOC once more; worse, that fresh `acquire()` captures the bumped epoch as its own baseline and *commits*. The loop now checks the epoch between the sleep and the next claim.
4. **Block-path error surface** — `ensureLease` swallowed the acquire failure, so `ensureHeldForUpload` mislabeled Bee outages/timeouts as "batch fully leased by other devices". The real cause now propagates.

One cross-coordinator race is intentionally **not** fixed here: `teardown()`'s detached `release()` sentinel can land after an immediate same-device re-auth re-claims the partition (lock SOC is last-write-wins), releasing the successor's claim to peers. That needs generation-fenced release semantics in `partition-lock.ts` — proposing it as a follow-up issue rather than growing this PR.

Issue #313, opened 2026-05-04
Consider having only identity level


agazso:
Currently there is an account level key, and each account can have multiple derived identities. This can be confusing for people who only want to create an identity and also confusing when using multiple identities, because the authorization is the same for the same account.

A major simplification could be to have only a single identity level, both from UX and code level. This would not affect the current developer API.

Issue #339, opened 2026-06-04
Restructure stored account data as nested account objects (account owns identities / apps / stamps)


vojtechsimetka:
## Summary

Restructure the locally-stored account data so that **the account is a single nested object that owns its identities, connected apps, and postage stamps** — instead of four independent flat collections joined by id pointers.

[#332](https://github.com/snaha/swarm-id/pull/332) explored the *behavioral* half of this (making the account an aggregate root in code, via a `SyncedAccount` facade) but deliberately left the storage shape flat. This issue is the *data-model* half: make the persisted shape itself nested, so the in-memory model matches the shape we already serialize everywhere data leaves the app.

This is about the data model and the reasoning for it — not about any single bug it happens to fix.

## Current model (flat, pointer-joined)

State lives in four independent localStorage collections, each managed by its own versioned storage manager and its own reactive store:

| Key | Shape | Relation to parent |
| --- | --- | --- |
| `swarm-id-accounts` | `Account[]` | root |
| `swarm-id-identities` | `Identity[]` | `identity.accountId` → account |
| `swarm-id-connected-apps` | `ConnectedApp[]` | `app.identityId` → identity |
| `swarm-id-postage-stamps` | `PostageStamp[]` | **no stored pointer** — account association is *re-derived* via `collectAccountStampBatchIds(account, identities)` |

Relationships are reconstructed at read time by filtering global arrays: `getIdentitiesByAccount(accountId)`, `getAppsByIdentityId(identityId)`, and the derived stamp-collection helper. There is no structural containment — the tree only exists implicitly, recomputed on every read.

## Why the flat shape is the wrong model

1. **The data is a tree, stored as disjoint tables.** Conceptually it's `account → identities → apps` plus account/identity-scoped stamps. Representing that as four flat arrays joined by foreign-key-style strings is an impedance mismatch we pay for on every read and every write.

2. **No referential integrity.** Because the links are pointers into separate arrays, nothing prevents orphans: an app whose identity was removed, or a stamp whose owning account is gone. Integrity is a convention enforced by call sites, not by the structure.

3. **Cross-cutting mutations must hand-cascade across four stores, in a careful order.** Account deletion / sign-out has to collect stamp batch ids *before* removing identities, precisely because the stamp↔account association is derived from the identity pointers it's about to destroy. That ordering constraint is a direct symptom of the model, and it's exactly the kind of thing that silently breaks.

[...]

PR #348, opened 2026-06-10
feat(ui): account lifecycle (create, sign-in, restore, manage) and dApp connect for the new identity UI

vojtechsimetka:
## What

Implements the account lifecycle in the new identity UI (`ui/`) following the SwarmID MVP Figma frames: creating, signing in to, restoring, managing, and connecting accounts — plus the dApp connect handshake and the `/proxy` iframe endpoint that make the new UI usable as a trusted domain end to end. Includes dev-environment housekeeping (deliberate ports, explicit new/legacy dev scripts).

### Flows

- **Create**: `/account/new` (name) → `/account/new/phrase` (secret recovery phrase: generate with reveal/copy/download, or import with validation) → `/account/new/access` (passkey / ETH wallet / password, with pending + cancel + error states) → `/account/new/done` (stamp upsell) → `/home`
- **Sign in**: `/account/import` — enter a phrase; switches to an existing local account or continues to access security
- **Restore**: `/account/restore` — `.swarmid` backup file + phrase; legacy-app exports are detected and explained rather than failing as "invalid file"
- **Manage** (`/home` Apps/Account tabs): rename, change unlock method, reveal private key/phrase, export backup, per-app disconnect/remove, delete account, account switcher, settings menu with theme preference
- **Connect**: `/connect` popup — pick an account, unlock, derive the per-app secret, and hand it to the dApp's proxy iframe via shared storage (or `setSecret` postMessage when storage is partitioned); `/proxy` serves the iframe endpoint

### Architecture decisions

- **Nested account model from the start** — the account is the aggregate root and owns its `stamps` and `connectedApps` directly as plain nested JSON under one localStorage key (`swarm-id-accounts-v2`). This is the model #332 approximates as a facade over flat stores in the legacy app, done natively here. No identity level — accounts only (#313).
- **The recovery phrase is the master secret.** Unlike the legacy app (passkey PRF / SIWE challenge as entropy), the 12-word BIP-39 phrase is generated or imported first, and the access-security step only chooses how the seed entropy is **encrypted at rest**:
  - passkey: WebAuthn PRF output → HKDF → AES-GCM (`userVerification: required` at registration *and* assertion — CTAP2 keeps separate PRF secrets per UV state)
  - ETH wallet: plain EIP-1193 `personal_sign` (no web3-onboard); key derived from the **deterministic message signature** (a secret only the wallet holder can produce) + stored salt, so re-signing the same message on any device unlocks the seed. Smart-contract wallets without recoverable ECDSA signatures are rejected up front.
  - password: PBKDF2 (600k iterations) + stored salt
- **Shared-storage lifecycle is symmetric.** The connect handshake upserts the shared `accounts`/`identities`/`postageStamps` records the proxy resolves connections and upload stamps against, and disconnect/remove/delete revoke them (secret dropped, sync tombstone set) so the dApp's proxy iframe actually de-authenticates.
- Storage keys are versioned and distinct from the legacy app's, since both UIs share an origin on GitHub Pages.
- Central `routes.ts` constants for all navigation; no raw route strings.

### Hardening (from a full-PR review pass)

- Unlock/ceremony dialogs use per-attempt tokens: cancelling an unabortable wallet/passkey prompt and retrying can no longer double-complete when the stale prompt is approved; `busy` guards cover the Enter-key paths.
- The setup draft survives back-navigation (renaming an account no longer regenerates the phrase behind an already-downloaded backup) and is cleared only after a successful connect handshake, so a failed handshake is retryable.
- Re-importing a phrase that already has a local account preserves its stamps, connected apps, and creation date instead of silently wiping them.
- Storage-partitioning detection remembers a consumed challenge per tab, so a popup reload/back-nav can't push a non-partitioned browser into download-only mode.
- `hexToBytes` rejects malformed input; backup encryption reuses the shared AES-GCM envelope + HKDF helpers; import/restore failure screens surface the real error instead of always blaming the phrase.

### Dev environment

- Ports moved off Vite's auto-increment range: new UI **5500** (preview 5501), legacy UI **5510**, demos **3500**/**3000**; pinned with `autoPort: false` in the preview launch config since every app is origin-sensitive.
- New stack scripts encode the generation pairing: `dev:new`, `dev:legacy` (= `dev`), `dev:ui:new`/`dev:ui:legacy`, `dev:demo:new`/`dev:demo:legacy`; CI builds `lib` before the new UI.

## Verification

- 24 unit tests: phrase↔entropy round-trips, known-vector address derivation, AES-GCM round-trips per access method, wallet-signature determinism, wrong-password rejection, backup round-trip + legacy/garbage-file rejection.
- Walked end-to-end in the browser (password path): create → back-nav rename (phrase preserved) → done → home → unlock → phrase reveal round-trips; account persists with the expected nested shape across reloads; screens validated against the Figma file.
- `@swarm-id/ui` format, lint (license headers), svelte-check, knip, and tests all green; root `check:all` green.

## Reviewer notes

- Passkey and ETH-wallet ceremonies need a real authenticator/wallet to exercise; the PRF-unsupported and smart-contract-wallet error paths were verified headless.
- "Add a postage stamp" routes to `/home` for now — the stamp purchase flow is still TBD in the design; stamps currently only enter via restored backups.
- New-UI `.swarmid` backups use a phrase-derived key and are intentionally not interchangeable with legacy exports (detected with a dedicated message).


agazso (approved):
The code is good and it works well, but it was a bit confusing that the title and description says "nested account model", but in fact this PR does not change the account model.

As a followup that (#339) can be done as a next step.

PR #359, opened 2026-06-16
feat(sync): intent SOCs resolve free-partition races across disjoint gateways

agazso:
## Problem

The per-partition lock SOC lives at a single static address. Across disjoint
gateways a node serves that address from a frozen cache (no TTL, no
invalidation), so two devices contending for a *free* partition never see each
other's lock and both bind it — two devices end up holding the same partition
(reproduced with 3 browsers on the public gateway). Tuning guard/TTL timeouts
can't help: the failure is non-visibility, not latency.

## Solution

Phase 2 of the partition-lock hardening (builds on #356's release-fencing,
already in main):

- **Intent SOCs** (`lib/src/sync/partition-intent.ts`): per-`(partition,
  deviceId, epoch)` address. Each contender writes only its own and reads every
  known peer's. The address rotates per epoch, so a read forces a real network
  retrieval that bypasses the frozen lock-SOC cache — contenders see each other,
  order by `generation`, and only the unique global-minimum binds the lock.
- **Deterministic home partition** (`keccak256(deviceId) mod K`): spreads
  devices across partitions so they rarely contend, and acts as the safety net
  when peers' locks are unreadable.
- **Reserved-slot routing**: an intent stamps into the contended partition's
  reserved slot (`< DATA_COUNTER_START`), so it can never overstamp user data.
- **Push-sync + diagnostics**: intent SOCs are receipt-backed; added
  `knownDevices` / `intentRound` claim diagnostics.

Plus dev tooling: split the dev-page reset into **Clear account** (account
stores only) and **Clear all** (every `swarm*` localStorage key + the
`swarm-utilization-store` IndexedDB DB, then reload).

## Residuals (documented in docs/Partition-Lock-Hardening-Plan.md)

- Absence is unprovable on Swarm: a timed-out intent read is treated as "no
  intent" to preserve liveness, so the guarantee is "when the network works,
  rivals are seen"; genuine network partition falls back to guard + TTL + refresh.
- The intent round is gated on the local (eventually-consistent) device
  registry, so a device not yet synced into a peer's registry can still be
  missed on a fresh claim (`freshClaim + no rival seen`). Flagged for follow-up.

## Test plan

- `pnpm --filter @snaha/swarm-id check:all` green — **753** unit + integration
  tests, incl. new `partition-intent.test.ts` and the expanded
  `partition-lease.integration.test.ts` (intent round, presence beacons,
  epoch-boundary, deconfliction backstop).
- Manual 3-device test (Firefox/Chrome/Brave, public gateway): no collisions
  once all three run current code.

## Notes

- Each package's `check:all` passes individually; the monolithic `pnpm
  check:all` wrapper flakes locally (vitest drops test files under full-suite
  load) — environmental, not a code issue.

### DIFF
--- a/.claude/rules/bee-cluster.md
+++ b/.claude/rules/bee-cluster.md
@@ -13,39 +13,35 @@ Docker-based local Bee cluster for development with postage stamps. Uses [@snaha
 pnpm dev:bee          # Start cluster (queen + 3 full workers = 4-node net), foreground
 pnpm dev:bee:detach   # Start in background
 pnpm dev:bee:stop     # Stop cluster
-pnpm dev:bee:fresh    # Fresh start (purge data); does NOT --pull (see below)
+pnpm dev:bee:fresh    # Fresh start (purge data)
 ```
 
 The scripts start a 4-node full network (`--full 4` = queen + 3 full workers).
 
-## CRITICAL: the Bee image must be built with `REACHABILITY_OVERRIDE_PUBLIC=true`
+## Pushsync needs `reachabilityOverridePublic=true` — now built into bee-compose ≥ 0.1.4
 
-`deferred: false` writes (every epoch-feed, partition-lock SOC, and any direct `/bytes`/`/chunks`/`/soc`
-upload) require Bee **pushsync** to complete a network **receipt**. A node only stores+receipts a chunk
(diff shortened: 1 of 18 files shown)


vojtechsimetka (approved):
All good and it works. I found one thing that could be improved and explored this direction with AI. Please have a look if this is worth addressing https://github.com/snaha/swarm-id/pull/366

PR #367, opened 2026-06-22
feat: nested single-level account model (#339/#313) + multi-device read/pull (#338)

agazso:
## Summary

Refactors account state to a **nested, single-level account model** and improves multi-device sync.
Resolves the account-architecture cluster #339 (nested data model) + #313 (collapse the account→identity
tiers into one) and #338 (read/pull triggers); partially addresses #337 (merge dedup landed; stamp
deletion tombstones still to come).

> **Draft + stacked.** This branch is built on top of `fix/intent-soc-gateway` (#359) and targets it as
> its base, so this diff is just the account-refactor work. Merge #359 first. Pre-production, so there is
> **no migration** — the nested model is the only model.

Design rationale and the running implementation log are checked in:
- `docs/Account-State-Refactor-337-338-339.md` (design + research)
- `docs/Account-State-Refactor-Implementation-Log.md` (what each commit landed)

## What changed

**Single-level account (#339 + #313).** The identity tier is gone. The account is the aggregate root and
the app-facing identity: `account.id` is the address apps connect to; per-app secrets derive from the
account master (`deriveSecret(master, appOrigin)`); the account private key (mnemonic-equivalent) stays
first-party, exposed only after explicit permission. The account **owns** its connected apps and postage
stamps as nested collections — one versioned localStorage document — instead of four flat collections
joined by `accountId`/`identityId` pointers. Independent keypairs = separate accounts (like Google
accounts). Multiple postage batches per account with one default + optional per-app override.

- **lib**: schema collapsed (no `Identity`); `ConnectedApp` keyed by `appUrl` with a `postageStampBatchID`
  override; `AccountStateSnapshot` drops `identities`; one nested storage manager; merge primitives
  (`mergeConnectedApps`/`mergePostageStamps`/`mergeDevicesList`) exported as the single source of truth;
  proxy reads the one nested doc.
- **swarm-ui**: `identity/[id]/*` → `account/[id]/*`; connect picks an **account**; home/drawer/app-list/
  stamps reworked to account scope; identity-create flows removed.
- **ui/** (new UI): connect handshake writes the single nested shared-account document.

**#338 read/pull triggers (swarm-ui).** Agent accounts now restore their published state from Swarm on
re-entry (2nd device adopts apps/stamps/devices instead of starting blank); `refreshAccountFromSwarm`
fires on app load / account switch (account `[id]` layout), not only the Devices view.

**Multi-device partition fixes** (concern B; fixing pre-existing `fix/intent-soc-gateway` code surfaced by
[...]

PR #373, opened 2026-06-24
feat(ui): add developer tools page to the new UI

vojtechsimetka:
Adds the developer tools page to the new `ui/` package, plus two supporting refactors it surfaced. Three commits, each self-contained.

## 1. Developer tools page (`/dev`)

Ports the legacy dev page into `ui/`, adapted to the new **nested single-level account model** (the `Account` aggregate owns its drives and connected apps inline — no separate identities/stamps/sync stores). Reachable at `http://localhost:5500/dev`. Two fully-working tabs, no mocks:

**Overview**
- Live status dots for the local Bee endpoints (Queen `:1633`, Worker `:16331`, Blockchain RPC `:9545`) + the demo origin.
- "Test connect flow" link to the demo (hash-encoded connect request, `:3500`).
- Local-data counts and a **Clear all data** action (also de-authenticates dApp proxy iframes via `removeAllSharedAccountRecords()`).

**Stamps**
- Buy a **mutable** postage stamp on the local chain (amount autofilled from chain price; mutable so the multi-device partition-lock SOC can be rewritten).
- List the node's stamps, showing which account holds each as a drive (and which is default).
- Assign a node stamp to an account as a **drive** (records the signer key, sets it default), with a custom-signer option and remove-from-account.

Dev-only code is colocated under `ui/src/routes/dev/` (`+page.svelte`, `bee.ts` Bee-node I/O helper, `copy-button.svelte`, `status-dot.svelte`).

_Intentionally omitted:_ the legacy **Sync** and **Devices** tabs — the new model has no identities or multi-device/sync subsystem wired in yet, so they'd be non-functional mock UI.

## 2. `Account` is a rich reactive object that owns its mutations

Per-account state changes are methods **on the account object** (`account.addDrive(drive)`, `account.rename(name)`, …) instead of store methods that take an id and look the account up.

- `Account` becomes a reactive class (`$state` fields) that persists itself via an injected `onChange`; `accountsStore` is now just the collection — `accounts`, `get(id)`, `add(record)` (returns the live instance), `remove(id)`, `clear()`.
- Types split in `$lib/types`: `AccountData` (portable, no secrets) and `AccountRecord` (data + access/encryptedSeed) are standalone interfaces; the `Account` class type is **type-only re-exported** so `$lib/types` stays the type entry point with no runtime cycle.
- The class carries a private `#onChange` (making it nominal), so `createBackup` — which only reads portable fields — takes `AccountData`.
- Call sites updated to the object form: `home-account`, `home-apps`, `connect-handshake`, account creation, dev page.

Reactivity confirmed: `$state` class fields re-render readers on mutation, independent of the `accounts` array.

## 3. Rename the account storage concept "stamp" → "drive"

"Drive" is the product term for an account's owned Swarm storage (backed by a postage stamp batch).

- **Domain → drive:** `PostageStamp`→`Drive`, `account.stamps`→`account.drives`, `defaultStampBatchId`→`defaultDriveBatchId`, `addStamp/removeStamp/setDefaultStamp`→`addDrive/removeDrive/setDefaultDrive`, plus UI copy (home "Drives" tab, "Add a drive", dev "Assign drive to account" / "N drives").
- **Kept as stamp (Bee-node layer / wire):** the dev page's buy-stamp & "Stamps on the Bee node" sections, `routes/dev/bee.ts` (`BeeStamp`/`fetchStamps`/`buyStamp`), the `/stamps` endpoint, bee-js `BatchId`, and the lib shared-record fields (`SharedPostageStamp`/`defaultPostageStampBatchID`/`postageStamps`).
- **Migration:** the persisted shape changed, so `accounts.svelte.ts` `load()` has a `normalize()` shim that accepts either spelling; existing local accounts survive and are rewritten in the new shape on the next persist.

## Verification

- `pnpm check:all` (full monorepo) passes — lint, `svelte-check` (0 errors/0 warnings for `@swarm-id/ui`), knip; 24 unit tests pass (incl. backup round-trip).
- End-to-end against a local `bee-compose` cluster: buy → list → assign → remove → clear; rich-object mutations persist and re-render; a **legacy-shaped** account migrates cleanly (loads as drives, rewrites in the new shape on mutation). No console errors.

> Note for reviewers: the new `ui` typechecks against `lib/dist`, so run `pnpm build:lib` after pulling if `svelte-check` reports phantom missing-export errors. The dev tab is still labeled "Stamps" since it's mostly Bee-stamp tooling.

### DIFF
--- a/ui/src/lib/components/home-account.svelte
+++ b/ui/src/lib/components/home-account.svelte
@@ -142,7 +142,7 @@
   function onNameChange() {
     const trimmed = name.trim()
     if (trimmed.length > 0 && trimmed !== account.name) {
-      accountsStore.rename(account.id, trimmed)
+      account.rename(trimmed)
     }
   }
 
@@ -268,7 +268,7 @@
       if (myAttempt !== attempt) {
         return
       }
-      accountsStore.setAccess(account.id, access, await encryptSeed(entropy, key))
+      account.setAccess(access, await encryptSeed(entropy, key))
       dialog = undefined
       newPassword = ''
       verifyNewPassword = ''
--- a/ui/src/lib/components/home-account.svelte
+++ b/ui/src/lib/components/home-account.svelte
@@ -104,7 +104,7 @@
   let toastMessage = $state<string | undefined>(undefined)
   let toastTimer: ReturnType<typeof setTimeout> | undefined
 
-  const isLocal = $derived(account.stamps.length === 0)
+  const isLocal = $derived(account.drives.length === 0)
   const accessLabel = $derived(methodLabel(account.access.type))
   const AccessIcon: Component = $derived(
     account.access.type === 'passkey'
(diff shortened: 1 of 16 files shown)


agazso (commented):
This PR seems to be a fork of the main branch before this PR was merged: https://github.com/snaha/swarm-id/pull/367

Therefore it redefines the `Account` type without a Zod type. I suggest to rebase on main first and then apply the changes from this PR.

PR #372, opened 2026-06-24
feat(sync): account-state — deletion tombstones (#337) + per-device feed sync (Phase 3)

agazso:
Continues the multi-device account-state refactor. Two threads in one PR.

## Phase 1 — deletion tombstones (#337) ✅ verified

Uniform deletion tombstones for the two account-held collections that lacked them (apps already had `revokedAt`):

- `PostageStamp.deletedAt`, `Device.removedAt`; tombstone-aware `mergePostageStamps`/`mergeDevicesList` (LWW, recency = `max(tombstone, activity)`); helper/proxy filtering; `removeStamp` tombstones + `removeDevice`; dev-page exercisers.
- 16 unit tests; verified end-to-end over real Swarm (local cluster + public gateway with an on-chain Gnosis batch): stamp delete & device removal propagate, removed device resurrects on re-sign-in.

Closes #337

## Phase 3 — per-device snapshot-feed sync (north star) 🚧 in progress

Replaces the single shared account-state snapshot feed (and its `verifyWon` retry dance) with one snapshot feed per device + a device-registry feed for discovery, merging on read with the Phase 1 primitives.

- Design + decisions: `docs/Account-State-Phase3-Design.md` (settled), log in `docs/Account-State-Phase3-Implementation-Log.md`.
- 3a commit 1 (additive, green): `device-state.ts`, `device-registry.ts`, `fold-account-from-swarm.ts` + the differential fold-equivalence gate.
- 3a commit 2 (cutover): switch the live write/read/proxy paths + retire the shared publish — next.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #374, opened 2026-06-24
feat(sync): multi-device account-state sync (tombstones + per-device feeds)

### LINKED ISSUE #337: Refactor: account-snapshot merge — dedup lib/UI logic + deletion tombstones for identities & stamps
The same account-snapshot merge semantics are implemented twice and can silently drift. Found during the #335 review.

## Problem

Two copies of the merge rules:

- **`lib/src/sync/merge-snapshot.ts`** — `mergeSnapshotWithRemote(local, remote)` (the **publish** path in `sync-account.ts`), operating on `AccountStateSnapshot`, with private helpers `mergeDevicesList` / `mergeIdentities` / `mergeConnectedApps` / `mergePostageStamps`.
- **`swarm-ui/src/lib/utils/refresh-account-from-swarm.ts`** — private `unionByKey` / `mergeConnectedAppsLww` / `mergeDeviceLists` (the **refresh** path, Devices tab), operating on store-shaped arrays and applying via the stores' `applyRefreshed` methods.

Both encode identical rules:
- union by natural key, **local wins** on collision (identities by `id`, stamps by `batchID`);
- connectedApps **last-writer-wins by `updatedAt`** (falling back to `lastConnectedAt`) **with revocation tombstones** (`revokedAt`);
- devices unioned by `deviceId`, preferring the larger `lastSignedInAt` (local on tie).

`refresh-account-from-swarm.ts:37-73` even comments that each helper "mirrors `merge-snapshot.ts`". If the rules change in one place (e.g. a new tombstone field, or a different tie-break) and not the other, publish and refresh diverge — a subtle multi-device correctness bug.

## Proposed fix

Make `lib/src/sync/merge-snapshot.ts` the single source of truth: export the small per-list primitives (a generic `unionByKey`, the connectedApps LWW+tombstone merge, the device union) and have `refresh-account-from-swarm.ts` import them instead of redefining.

Wrinkle: the two call sites operate on **different shapes** — `mergeSnapshotWithRemote` takes a whole `AccountStateSnapshot`, while the refresh path works on store-shaped arrays and applies them to stores. So the clean shared unit is the **per-list primitives**, not `mergeSnapshotWithRemote` wholesale. (Alternatively, refresh could build a local `AccountStateSnapshot`, call `mergeSnapshotWithRemote`, then apply the result to stores — heavier, but fully removes the second code path.)

## Relationship to other work

Independent of #336 (BatchWriteCoordinator). They only meet at #336's phase-5 *proxy-side publish*, which would **consume** this merge logic (build snapshot → merge → write via the coordinator) but doesn't own or move it. So this can be done on its own, in any order.

Refs: #335 (introduced the duplication), #336 (consumes it at phase 5), #217 (umbrella).

## Also: deletion tombstones for identities & stamps

`connectedApps` propagate removals across devices via tombstones (`revokedAt`); `identities` and `postageStamps` still **cannot** express a deletion — the union merge re-adds a locally removed entry on the next sync. Since this is the same merge layer, extend the tombstone pattern to identities and stamps here (or document why they intentionally differ).

## Acceptance

- One definition of the merge rules, imported by both publish and refresh.
- Refresh + publish behaviour unchanged.
- A shared unit test covers the LWW/tombstone + device-union cases once (today only `merge-snapshot.test.ts` exercises them).
- Identities and stamps propagate deletions across devices (or a documented decision not to).
- `pnpm check:all` green.

agazso:
Multi-device account-state sync for Swarm-backed identities. Each device owns its state; convergence happens on read. Supersedes the single shared snapshot feed + `verifyWon` write dance.

(Branch renamed from `feat/account-deletion-tombstones`; the GitHub rename API closed the original #372, so this PR supersedes it. `Closes #337` carried over.)

**Deletion tombstones (#337).** `PostageStamp.deletedAt` + `Device.removedAt` (apps already had `revokedAt`); tombstone-aware `mergePostageStamps`/`mergeDevicesList` (LWW, recency = `max(tombstone, activity)`); `removeStamp`/`removeDevice`. A removed device resurrects on re-sign-in.

**Per-device snapshot feeds (Phase 3a).** Each device publishes its full view to its OWN feed (`device-state.ts`); `foldAccount` merges all devices on read with the Phase 1 primitives — no shared SOC address, no cross-device write contention. Cutover retires the shared publish path.

**Append-only device roster.** Discovery moved from a shared mutable registry (clobbered on a caching gateway) to an append-only roster (`device-roster.ts`): each device only ever appends its own record, so a fresh device can't drop an existing one. Read with a windowed-parallel scan.

**Per-field scalar clocks.** `accountName` / `defaultPostageStampBatchID` / `settings` carry their own LWW clocks (`*At`), so a peer's change to one scalar propagates on refresh without clobbering a concurrent change to another.

**Read-path parallelization.** `foldAccountFromSwarm` reads all device feeds under one `Promise.all` and the roster scans in parallel windows — gateway fold dropped ~2.3× (3 devices: avg 5.9s → 2.6s).

Design + log: `docs/Account-State-Phase3-Design.md`, `docs/Account-State-Phase3-Implementation-Log.md`. Verified end-to-end on the public gateway with an on-chain Gnosis batch (3-device convergence, tombstone propagation, device resurrection, partition contention).

Closes #337


agazso:
Post-review fix (68f21ff3): a critical pass surfaced one latent defect — a deleted stamp could never be re-added. `mergePostageStamps` recency was `deletedAt ?? createdAt` (re-add could never beat its tombstone) and `addStamp` threw "already exists" for a UI-invisible tombstoned record. Recency is now `max(deletedAt, createdAt)` (mirrors the device clock) and `addStamp` resurrects a tombstoned batch instead of throwing. Backward-compatible — existing tombstone-wins tests still pass; added a resurrection test. Latent today since stamp deletion isn't user-reachable yet, so this guards `removeStamp` for when it ships.

vojtechsimetka (approved):

PR #376, opened 2026-06-25
feat(ui): developer tools page on a single Zod-validated account model

vojtechsimetka:
Supersedes #373. Ports the developer tools page into the new `ui/` package **and** resolves that PR's review (agazso): _"redefines the `Account` type without a Zod type."_ The product UI now runs on the shared `@snaha/swarm-id` Zod `Account` — a single validated source of truth — while keeping the rich account-object ergonomics #373 introduced.

## 1. Developer tools page (`/dev`)

Reachable at `http://localhost:5500/dev`. Built on the **single nested account model** (the account owns its drives and connected apps inline). Four working tabs — including **Sync** and **Devices**, which #373 intentionally omitted because the old model had no sync subsystem wired in:

**Overview**
- Live status dots for the local Bee endpoints (Queen `:1633`, Worker `:16331`, Blockchain RPC `:9545`) + the demo origin.
- "Test connect flow" link to the demo (`:3500`).
- Local-data counts and a **Clear all data** action.

**Stamps**
- Buy a **mutable** postage stamp on the local chain (amount autofilled from chain price).
- List the node's stamps, showing which account holds each as a drive (and which is default).
- **Assign drive to account** (records the signer key, sets it default), with a custom-signer option and remove.

**Sync**
- **Sync All Accounts** — publishes each account's state snapshot to Swarm (epoch-feed SOC + partition lease).

**Devices**
- Per-account device list with **refresh-from-Swarm** (reads the published snapshot back) and lease/partition debug state.

Dev-only code lives under `ui/src/lib/dev/*` and `ui/src/routes/dev/*`.

## 2. One Zod-validated account model (the review fix)

#373 made `Account` a hand-rolled reactive class with **no Zod** and kept **two** account stores bridged by `mirror.ts`. This collapses both into the shared model:

- **Extend the lib schema** rather than reuse-as-is: add a `local` variant to `AccountSchemaV1` (`LocalAccountSchemaV1` + `AccessMethodSchemaV1` — passkey / eth-wallet / password + `encryptedSeed`), so the UI's existing auth is representable **with no crypto/behavior change**. The existing `passkey`/`ethereum`/`agent` variants are untouched; the UI's three methods nest under `account.access` so they never collide with the top-level `type` discriminator.
- `local` branches added to `serializeAccount` and `buildBackupHeader` (`LocalBackupHeaderSchemaV1`); new schemas/types exported. New unit tests cover the `local` serialize round-trip + backup header.
- **Single reactive store** on the lib key `swarm-id-accounts` — the SAME document the proxy iframe and sync engine read. Product UI, `/dev`, and sync all drive off it; `mirror.ts` and the parallel store are **deleted**. `/dev` changes now surface on Home immediately (no mirroring).
- Account creation computes `derivationKey` from the master key (entropy in hand); ids/keys render `0x`-prefixed via a `display0x` helper (bee-js `.toHex()` is bare).

## 3. `Account` is a rich object that owns its mutations

Kept from #373's commit 2, now wrapping the validated lib `LocalAccount`:

- `Account` is a reactive class (`$state` fields) — mutators are methods **on the object** (`account.rename(name)`, `account.addDrive(drive)`, `account.connectApp(app)`, `account.setDefaultDrive(id)`, …), persisting via an injected `onChange` (lib storage manager + optional sync hook). A private `#onChange` makes the class nominal.
- `accountsStore` is just the collection: `accounts` / `get` / `getAccount` / `add` (returns the live instance) / `remove` / `clear` / `reload`.

## 4. "stamp" → "drive" rename

"Drive" is the product term for an account's owned Swarm storage (backed by a postage-stamp batch):

- **Drive** (product): Home "Drives" tab, "Add a drive", `/dev` "Assign drive to account" / "N drives" / "Set/Remove account drive", and the account methods `addDrive`/`removeDrive`/`setDefaultDrive`/`updateDriveUtilization`.
- **Stamp** (Bee-node / wire — unchanged): buy-stamp, "Existing Stamps (Bee Node)", `routes/dev/bee.ts`, the `/stamps` endpoint, bee-js `BatchId`, and the lib fields `postageStamps` / `defaultPostageStampBatchID` (a drive _is_ a stamp batch).

Also narrows two legacy `swarm-ui` call sites for the widened union.

## Verification

- **`pnpm check:all` passes** — lint, `svelte-check` (0 errors), knip, 790 lib + ui tests (incl. the new `local` serialize round-trip and backup round-trip).
- **End-to-end against a fresh local `bee-compose` cluster** (queen + 3 workers + blockchain, reachability `Public`, non-deferred upload ~0.1s):
  - Create account → `type:'local'` record (bare-hex id, 64-char `derivationKey`, `access`, `encryptedSeed`, key `swarm-id-accounts`; no legacy `swarm-id-accounts-v2`).
  - `account.rename()` → switcher re-renders from the `$state` field **and** persists.
  - `/dev` buy stamp (chain tx) → **assign drive** (`addDrive`/`setDefaultDrive`) → Home drops the view-only banner immediately (one store, no mirror).
  - **Connect** demo → `completeConnect()` → `account.connectApp()` writes the connected app **with a live `appSecret`** into the same record; `account.disconnectApp()` clears it.
  - **Sync All Accounts → "1 succeeded, 0 failed"**; **Devices → refresh-from-Swarm** reads the snapshot back.

> **Notes for reviewers**
> - `ui` typechecks against `lib/dist`; run `pnpm build:lib` after pulling if `svelte-check` reports phantom missing-export errors.
> - **Local sync testing** needs two env things (not code): a freshly-meshed cluster (`pnpm dev:bee:fresh` on `@snaha/bee-compose` ≥ 0.1.5 so node reachability is `Public`), and the Bee node URL set to `http://localhost:1633` in **Network settings** — sync defaults to the public gateway, which a localhost browser can't reach.
> - **Migration:** the lib key `swarm-id-accounts` is canonical; per PoC policy there is **no migration** of old `swarm-id-accounts-v2` records — a pre-existing local-only account would be re-imported from its phrase.

### DIFF
--- a/lib/src/schemas.ts
+++ b/lib/src/schemas.ts
@@ -243,12 +243,50 @@ export const AgentAccountSchemaV1 = CommonAccountSchemaV1.extend({
 })
 
 /**
+ * Access Method Schema V1
+ *
+ * How a `local` account's encrypted seed is unlocked on this device. Mirrors
+ * the identity UI's access methods. Fields are device-local crypto material
+ * (plain hex / numbers) — never bee-js byte classes — so they round-trip as-is.
+ */
+export const AccessMethodSchemaV1 = z.discriminatedUnion("type", [
+  z.object({ type: z.literal("passkey"), credentialId: z.string() }),
+  z.object({
+    type: z.literal("eth-wallet"),
+    walletAddress: z.string(),
+    encryptionSalt: z.string(),
+  }),
+  z.object({
+    type: z.literal("password"),
+    kdfSalt: z.string(),
+    kdfIterations: z.number(),
+  }),
+])
+
+/**
+ * Local Account Schema V1
+ *
+ * A device-local account whose BIP-39 entropy is encrypted at rest with a
+ * device-local access method (passkey / eth-wallet / password). This is the
+ * model the identity UI (`@swarm-id/ui`) creates and unlocks. Like every
+ * account it carries a `derivationKey` (computed from the master key at
+ * creation time); the seed itself is held only as `encryptedSeed`.
+ */
+export const LocalAccountSchemaV1 = CommonAccountSchemaV1.extend({
+  type: z.literal("local"),
+  access: AccessMethodSchemaV1,
+  // BIP-39 entropy encrypted with the access-method key (hex: IV || ciphertext).
+  encryptedSeed: z.string(),
+})
+
+/**
  * Account Schema V1 (discriminated union)
  */
 export const AccountSchemaV1 = z.discriminatedUnion("type", [
   PasskeyAccountSchemaV1,
   EthereumAccountSchemaV1,
   AgentAccountSchemaV1,
+  LocalAccountSchemaV1,
 ])
 
 // ============================================================================
@@ -308,6 +346,8 @@ export type Device = z.infer<typeof DeviceSchemaV1>
 export type PasskeyAccount = z.infer<typeof PasskeyAccountSchemaV1>
 export type EthereumAccount = z.infer<typeof EthereumAccountSchemaV1>
 export type AgentAccount = z.infer<typeof AgentAccountSchemaV1>
+export type LocalAccount = z.infer<typeof LocalAccountSchemaV1>
+export type AccessMethod = z.infer<typeof AccessMethodSchemaV1>
 export type Account = z.infer<typeof AccountSchemaV1>
 export type ConnectedApp = z.infer<typeof ConnectedAppSchemaV1>
 export type PostageStamp = z.infer<typeof PostageStampSchemaV1>
--- a/lib/src/index.ts
+++ b/lib/src/index.ts
@@ -170,6 +170,7 @@ export {
   PasskeyBackupHeaderSchemaV1,
   EthereumBackupHeaderSchemaV1,
   AgentBackupHeaderSchemaV1,
+  LocalBackupHeaderSchemaV1,
   EncryptedSwarmIdExportSchemaV1,
 } from "./utils/backup-encryption"
 
@@ -177,6 +178,7 @@ export type {
   PasskeyBackupHeader,
   EthereumBackupHeader,
   AgentBackupHeader,
+  LocalBackupHeader,
   EncryptedSwarmIdExport,
   BackupHeaderWithoutCiphertext,
   ParseHeaderResult,
@@ -326,6 +328,8 @@ export type {
   PasskeyAccount,
   EthereumAccount,
   AgentAccount,
+  LocalAccount,
+  AccessMethod,
   ConnectedApp,
   PostageStamp,
   AccountMetadata,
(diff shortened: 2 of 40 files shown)


lib/src/schemas.ts
```diff
@@ -242,13 +242,51 @@ export const AgentAccountSchemaV1 = CommonAccountSchemaV1.extend({
...
+/**
+ * Local Account Schema V1
+ *
+ * A device-local account whose BIP-39 entropy is encrypted at rest with a
+ * device-local access method (passkey / eth-wallet / password). This is the
+ * model the identity UI (`@swarm-id/ui`) creates and unlocks. Like every
+ * account it carries a `derivationKey` (computed from the master key at
+ * creation time); the seed itself is held only as `encryptedSeed`.
+ */
+export const LocalAccountSchemaV1 = CommonAccountSchemaV1.extend({
```
agazso, line 275: I don't think this is the right way to model this. In the new model we don't discriminate between local and synced accounts, an account is considered modal if it does not have a valid postage batch. Otherwise the rest of the fields are the same.

agazso (commented):
Merging this on top of #374 would compile without errors, but it would break some of the functionality implemented (and tested) there:

1. Read/refresh is dead. `refresh-account-from-swarm.ts` reads the shared `ACCOUNT_SYNC_TOPIC_PREFIX:<id>` feed via `AsyncEpochFinder.findAt`. On #374 that feed is no longer written (the constant is kept only so a test can assert it's retired); publishing goes to per-device feeds. So refresh always returns no-backup → /dev Devices "refresh-from-Swarm" and the peer list stop working. The fix is to delete those 124 lines and call #374's `foldAccountFromSwarm({ bee, derivationKey, accountId })`, which does the roster scan + per-device read + merge internally.
2. Drive deletion doesn't stick. `Account.removeDrive()` hard-filters the stamp out of the array. #374 closed #337 by adding a deletedAt tombstone and making mergePostageStamps tombstone-aware (recency = max(deletedAt, createdAt)). A hard delete with no tombstone gets re-added on the next fold from any device feed that still has it. #376's own comment ("no stamp tombstone yet — #337 — so a peer's union merge re-adds it") is exactly the thing #374 fixes — but #376 doesn't call the new tombstone path.
3. Scalar sync regresses a tested #374 feature. This is my original finding #1, now sharper: #374 added per-field clocks (accountNameAt/defaultStampAt/settingsAt) and a rename-clobber.test.ts to prove name/default-drive/settings converge across devices. #376's Account class doesn't carry those fields and never bumps lastModified, so sync-account.ts:282 publishes account.accountNameAt ?? createdAt = frozen createdAt. Rebased, #376 would un-fix rename-clobber. The mutators need to stamp the matching *At, and applyRefreshed must fold scalar metadata back (which foldAccountFromSwarm returns).

PR #377, opened 2026-06-26
refactor(lib)!: unify the account model — one BIP-39 seed account

vojtechsimetka:
First chunk of breaking #376 into reviewable pieces. It **unifies the account model into a single account type**, and — because the legacy UI was built around the old multi-type model — **migrates `swarm-ui` to use BIP-39 seed-phrase accounts** as a consequence (with minimal UI changes, just enough to keep it working).  @agazso This also resolves your account-modeling concern by removing the account-type distinction rather than adding more variants.

## Unifying the account type

The three former account types — `passkey`, `ethereum`, and `agent` — shared the same cryptographic core. Each is a BIP-39 seed account that derives its master key, its `derivationKey`, and every sub-key through one identical pipeline. What actually differed between them was a single concern: **how the seed is protected at rest and recovered on a device** — via a passkey's PRF output, a wallet signature, or manual re-entry of the recovery phrase. The `type` discriminant elevated that protection method to a top-level entity distinction, even though the sync engine and key derivation never branch on it — they operate purely on `derivationKey` and the postage batch.

This refactor models the protection method where it belongs instead of as an account type:

- **One `Account`** — a single object with no `type` discriminant. The seed-protection method is an optional `access` (passkey | eth-wallet | password) paired with an `encryptedSeed` vault; both are present on an interactive UI account and both absent on a headless/programmatic one.
- **`isLocalAccount(account)`** — "local vs synced" was likewise not a type but a runtime capability: whether the account owns a usable, on-chain, non-tombstoned postage batch (and can therefore pay to upload/sync). It is now a derived predicate, not a stored field.

## Migrating the legacy UI to BIP-39 seed-phrase accounts

`swarm-ui` was built around the old `passkey`/`ethereum`/`agent` types and branched on `account.type` throughout, so unifying the model requires bringing it onto the seed-phrase account too. Scope is deliberately minimal — `swarm-ui` is being superseded by `ui/`, so this is just enough to keep it compiling and its core flows working:

- Every account is now a BIP-39 seed account; each create flow secures the seed with an access method (passkey / wallet / password) instead of producing a distinct account type.
- Auth unlocks via the access method and decrypts `encryptedSeed`; backup import is type-agnostic; the drawer renders by `access.type`.
- Duplicated the seed/access crypto from `ui/` into `swarm-ui` (no cross-package dependency) and removed the now-dead ethereum/passkey/secret-seed modules + `@web3-onboard`.

## Changes

- **`refactor(lib)!:`** remove the `passkey`/`ethereum`/`agent` variants, the discriminated union, and `type`; `serializeAccount` and the backup header de-branch (one header, no `accountType`, no key material); one `createAccount` test fixture.
- **`refactor(swarm-ui):`** the legacy-UI migration described above.

Net **−1,900 lines**.

> [!NOTE]
> swarm-ui behavior change (acceptable for a retiring UI): passkey/wallet no longer reconstruct an account on a fresh device — they only encrypt a locally-stored seed — so sign-in unlocks local accounts or imports a `.swarmid` phrase backup. The eth secret-seed / generation-details screens were removed.

## Verification

`pnpm check:all` passes — prettier, eslint (license headers), svelte-check on both UIs (0/0), knip, and 812/812 lib tests. (swarm-ui verified compile/type/lint/test; its interactive passkey/wallet flows were not runtime-tested.)


lib/src/schemas.ts
```diff
@@ -193,14 +216,23 @@ export const PostageStampSchemaV1 = z.object({
...
   publicKey: CompressedPublicKeySchema.optional(),
+  // Device-local seed vault. Every identity-UI account is a BIP-39 seed account:
+  // `encryptedSeed` holds that seed encrypted at rest and `access` records the
+  // method used to encrypt/unlock it on this device (passkey / eth-wallet /
+  // password). The two travel together (both present on every UI account, both
+  // absent on legacy `swarm-ui` accounts). Device-local: deliberately excluded
+  // from the sync snapshot.
+  access: AccessMethodSchemaV1.optional(),
+  // BIP-39 entropy encrypted with the access-method key (hex: IV || ciphertext).
+  encryptedSeed: z.string().optional(),
```
agazso: These should be not optional, they should always be required.

lib/src/schemas.ts
```diff
@@ -182,7 +174,38 @@ export const PostageStampSchemaV1 = z.object({
...
+ * never bee-js byte classes — so they round-trip as-is. Paired with
+ * `encryptedSeed` on the account: the two travel together (both present on an
+ * interactive UI account; both absent on a headless/programmatic account that
+ * holds its seed in memory).
+ */
+export const AccessMethodSchemaV1 = z.discriminatedUnion("type", [
+  z.object({ type: z.literal("passkey"), credentialId: z.string() }),
+  z.object({
+    type: z.literal("eth-wallet"),
+    walletAddress: z.string(),
```
agazso: This is not necessary, the `account.id` is the ethereum address of the private key anyways

agazso:
I would keep the `@web3-onboard` package, it allows you to select a wallet when there are multiple wallets installed in the browser. For example Brave has a built-in, but you can also install a custom one.

lib/src/schemas.ts
```diff
@@ -325,15 +320,33 @@ export const AccountStateSnapshotSchemaV1 = z.object({
...
+ * for uploads, so it is effectively view-only and has nothing to sync. "Local"
+ * is purely this runtime predicate; it is NOT a stored field or account `type`.
+ *
+ * A stamp counts as a valid drive when it `exists` on-chain and is `usable`
+ * (synced enough to upload against) and is not tombstoned. A freshly
+ * purchased-but-not-yet-usable stamp therefore still reads as local until it
+ * becomes usable.
+ */
+export function isLocalAccount(account: Account): boolean {
+  return !account.postageStamps.some(
```
agazso: This should check if validity of the `defaultPostageStampBatchID` , not all stamps

lib/src/schemas.ts
```diff
@@ -325,15 +320,40 @@ export const AccountStateSnapshotSchemaV1 = z.object({
...
+ */
+export function isLocalAccount(account: Account): boolean {
+  const batchID = account.defaultPostageStampBatchID
+  if (batchID === undefined) return true
+  const stamp = account.postageStamps.find((s) => s.batchID.equals(batchID))
+  return !(
+    stamp !== undefined &&
+    stamp.exists &&
+    stamp.usable &&
+    stamp.deletedAt === undefined
```
agazso, line 355: This should also check if the stamp has not expired yet. This can be done separately though, because it should be first clarified what is the user experience when a batch expires.

lib/src/schemas.ts
```diff
@@ -193,14 +218,21 @@ export const PostageStampSchemaV1 = z.object({
...
   // App-facing public key (compressed secp256k1). Surfaced to dApps as the
   // identity public key in ConnectionInfo. Was previously on the identity.
   publicKey: CompressedPublicKeySchema.optional(),
+  // Device-local seed vault. Every account is a BIP-39 seed account: `access`
+  // records how the seed is protected/unlocked on this device (passkey /
+  // eth-wallet / password) and `encryptedSeed` is that seed encrypted at rest.
+  // Device-local: deliberately excluded from the sync snapshot.
+  access: AccessMethodSchemaV1,
+  // BIP-39 entropy encrypted with the access-method key (hex: IV || ciphertext).
+  encryptedSeed: z.string(),
```
agazso: The encryptedSeed should not allow empty values

```suggestion
  encryptedSeed: z.string().length(64), // 64-char hex
```

swarm-ui/src/lib/seed-phrase.ts
agazso, line 1: This file no longer seems to be relevant and the comment added to it seems to be incorrect. There is no account without access/encryptedSeed. The best candidate for the agentic account is an account with password access type.

agazso (approved):
I tested it, works well.

There seems to be leftovers in comment and code from a state where the encryptedSeed was optional for agentic accounts. I would simply use the regular password access method for agents.

PR #382, opened 2026-06-30
feat(sync): crash-safe commit-ordered partition handoff + fast acquire/upload

agazso:
## Summary

Implements `docs/Partition-Acquire-Optimistic-Lease.md` and fixes the multi-second partition **acquire + SOC upload** latency seen in the demo, makes the per-upload publish cheap and crash-safe, and adds a live (real-backend) test suite that exercises the full partition handoff machinery end to end.

On a freshly-reloaded page the demo went from ~40s acquire + ~40s upload to a sub-second held-lease upload; the remaining cold-acquire cost is now dominated only by the public gateway's slow 404s, not by in-library work.

### 1. Crash-safe commit-ordered handoff (ack-after-publish) — correctness
- `BatchWriteCoordinator.withWrite` publishes the partition-state counter (`PartitionLease.publishState`) **before** resolving an upload, so an un-acked in-flight write resumes at exactly the published counter on an orderly takeover — a crash/lapse between the data write and the publish loses only the unacked write (content-addressed → safe retry), never an acked one.
- Reverse-clobber guard: a post-op lease re-validation keyed on `leasedUntil - LEASE_SKEW_MARGIN_MS` refuses to ack/publish when the lease could have lapsed. It is **throttled to zero reads on the fast path** — a healthy fast upload publishes without re-reading the lock SOC, so the guarantee covers the same-gateway lease-lapse case; the data write itself is unfenced and the disjoint-gateway dual-hold remains a documented residual (see Postage-Batch-Partitioning.md §12), resolved by the refresh-tick beacon backstop rather than this guard.

### 2. Rotating per-epoch state-pointer SOC (replaces the walked epoch feed) — latency
- The pointer to the latest counter-reference chunk is published at a deterministic, time-bucketed SOC address (mirrors the occupancy beacon). The reader **computes** the address and reads the current/previous bucket (and the gone holder's `leasedUntil` bucket) directly — no 32-level epoch-tree walk (~27 slow gateway 404 probes, the dominant pre-upload stall). The rotating address also bypasses a gateway's frozen static cache.
- Drops the epoch-feed finder/updater, `EpochUpdateHints` plumbing, and the +1 feed-SOC slot compensation (the pointer is a reserved-slot SOC).

### 3. Incremental + parallel partition-state publish
- `writePartitionState` re-uploads only the counter chunks that changed since the last publish (diffed against the previous published counter, reusing retained refs) and fires all PUTs in a single `Promise.all`. Steady-state publish drops from ~34 serial uploads to a few in one round-trip. Zero counter chunks use an all-zero sentinel ref (sparse publish).

### 4. Incremental **first** publish after acquire (fix)
- `claimPartition` now seeds the lease's `publishedReferences`/`publishedCounter` from the state read at acquire, so the **first** publish of a session is incremental (only the changed chunk + reference chunk) instead of re-publishing every non-zero chunk. `readPartitionState` returns the per-chunk refs (full read, and best-effort on a cache hit) to enable this; a genuinely fresh account still falls back to the sparse-full path.

### 5. Acquire-path latency fixes
- **Non-blocking device-registry refresh:** `acquire()` no longer awaits `refreshKnownDeviceIds()` under the cross-tab write lock (previously stalled the first upload up to the 45s acquire cap).
- **Roster read timeouts:** each SOC + blob read bounded by a 2.5s race.
- **`foldAccountFromSwarm` in-flight coalescing:** concurrent folds for the same account share one promise, collapsing 3-4 uncoordinated UI folds into one read set.
- **`tryCreateTag` caches the "tags unsupported" 404 per `Bee`** so every upload no longer pays that blocking round-trip on public gateways.

### 6. Live (remote-backend) test suite — `lib/test/live/`
Renamed from `lib/test/multi-device/` (it now holds single-device tests too, and "multi-device" clashed with the unrelated product concept). Opt-in, never in CI, runs against a live Bee backend (default: public gateway) with a real funded postage batch; config `vitest.live.config.ts`, script `pnpm test:live`.

- `single-device-acquire-upload` — clean acquire → upload → publish wall times.
- `multi-device-acquire-upload` — 2 devices, per-device acquire + upload + publish timings.
- `three-device-acquire-handoff` — full p1 handoff **cycle B → C → B** (A holds p0 throughout): cold acquire, intent round, expired-holder takeover in both directions, and cross-device state resume.
- `second-upload-delay`, `upload-cost-breakdown` — held-lease re-validation throttle and upload phase breakdown.

## Tests
- Unit + integration: partition-state/partition-lease integration rewritten for the pointer transport (seed pointers, no +1 compensation, incremental round-trip, `leasedUntil`-bucket lookup, ack-after-publish ordering, reverse-clobber guard, seeded incremental first publish); new unit tests for fold coalescing, the tags cache, and acquire not blocking on a slow refresh.
- Full `@snaha/swarm-id` suite passes; typecheck/lint/format clean. The only intermittent failure is the pre-existing flaky "disjoint slots" stamper test (unrelated; passes in isolation).
- Live suite verified against the public gateway, including the B→C→B handoff cycle.


vojtechsimetka (approved):
Approving — the correctness core held up under review: ack-after-publish ordering, the incremental-diff-against-published-baseline, retained-chunk collision handling (`forceFull` re-pin + fail-safe read-only degradation), timer lifecycle on lease replacement, and fold-coalescing eviction were each attacked and came back clean.

A few things to address, mainly code duplication:

**1. `partition-state.ts:790` — KEYSTONE monotonicity tripwire is silently skipped on the sparse-full path.**
The guard is `if (incremental && previousCounter !== undefined)`, but `incremental` (line 773) already requires `previousReferences !== undefined && previousCounter !== undefined` — the second conjunct is dead code, and the tripwire never runs when `previousCounter` is known but `previousReferences` is missing. That state is reachable: on a cache-hit read the best-effort reference-chunk download can fail, seeding the lease with a counter but no refs. The tripwire exists to fail loud on exactly the broken-persistence case; here a regressed counter would publish silently. Fix is one line: guard on `previousCounter !== undefined` alone.

**2. Duplicated timing constants across four sync modules.**
The same 2.5s read timeout is defined independently four times — `ROSTER_READ_TIMEOUT_MS` (device-roster.ts:63), `INTENT_READ_TIMEOUT_MS` (partition-intent.ts:110), `LOCK_READ_TIMEOUT_MS` (partition-lease.ts:92), `STATE_POINTER_READ_TIMEOUT_MS` (partition-state.ts:153) — and `INTENT_EPOCH_MS = 30_000` (partition-intent.ts:101) duplicates `LEASE_TTL_MS` with a comment admitting they're "intended to track each other" but can't share a source because of a circular import through batch-utilization. A tiny dependency-free `sync/timing-constants.ts` would fix both the drift risk and the circular-import excuse. The ~20-line `withTimeout` + `TimeoutError`-classification + sentinel boilerplate repeated in device-roster.ts and partition-state.ts could fold into the same cleanup.

**3. `makeInMemoryCache()` is byte-identical in all five live test files** (single-device-acquire-upload.test.ts:51, multi-device-acquire-upload.test.ts:53, three-device-acquire-handoff.test.ts:63, second-upload-delay.test.ts:56, upload-cost-breakdown.test.ts:45). The suite already has shared helpers in `lib/test/live/env.ts` / `setup.ts` — move it there; five copies of a `UtilizationStoreDB` mock will drift the first time that interface changes.

**Minor notes (no action required):**
- `lib/vitest.config.ts:21` raises the global `testTimeout` to 30s for the whole unit suite; scoping it to the heavy crypto/handoff tests would keep the fast suite honest against future perf regressions.
- In `fold-account-from-swarm.ts`, callers that coalesce onto an in-flight fold share its rejection (next call after settle retries fresh) — correct, but worth a one-line comment since one slow-gateway failure fans out to every coalesced UI caller.

agazso:
Addressed in 0dbb5595:

1. **Tripwire on the full path** — guarded on `previousCounter !== undefined` alone (dropped the dead `incremental` conjunct), so the sparse-full path (cache-hit read whose ref download failed → counter known, refs absent) now fails loud on a regressed counter. Added a `partition-state.test.ts` case for it (red before, green after).
2. **Timing constants** — new dependency-free `sync/timing-constants.ts` holds `SYNC_READ_TIMEOUT_MS`, `LEASE_TTL_MS`, and `INTENT_EPOCH_MS = LEASE_TTL_MS`; the four read-timeout literals and the epoch/TTL pair now resolve to it, re-exported under their existing public names (no call-site churn, "circular import" excuse gone).
3. **`makeInMemoryCache`** — moved into `test/live/env.ts`; the five live tests import it.

Left as-is per your "no action required" notes: the global `testTimeout` scoping and the extra fold-account rejection comment. Also skipped the optional `withTimeout`+sentinel boilerplate fold — it's a thin abstraction over two call sites with different return shapes, so I don't think the indirection pays for itself, but happy to do it if you'd prefer.

`pnpm check:all` green.

PR #397, opened 2026-07-03
feat(ui): confirm before removing a drive

### LINKED ISSUE #390: New UI: removing a drive should ask for confirmation
In the Storage tab, the drive menu's **Remove** action removes the drive immediately, with no confirmation step (`remove()` in `ui/src/lib/components/home-drives.svelte` calls `account.removeStamp` directly).

Removal only drops the local stamp record (the batch stays on chain and can be re-attached via its Batch ID), but from the user's perspective it looks destructive — and the removal tombstone syncs to other devices. Add a confirmation dialog before removing.

Reported in the review of #375: https://github.com/snaha/swarm-id/pull/375#issuecomment-4868308983

vojtechsimetka:
The Storage tab's **Remove** action detached a drive immediately with no confirmation, even though removal looks destructive and the tombstone syncs to other devices.

This adds `drive-remove-dialog.svelte` (following the sibling drive-dialog pattern): the menu action now opens a confirmation dialog titled with the drive's name, explaining that removal only detaches the drive from the account — the batch stays on the Swarm network and can be re-attached later via its Batch ID. A destructive **Remove drive** button confirms; **Cancel** (or Escape/X) closes without action.

Closes #390

### DIFF
--- a/ui/src/lib/components/drive-remove-dialog.svelte
+++ b/ui/src/lib/components/drive-remove-dialog.svelte
@@ -1,0 +1,42 @@
+<!--
+  Copyright 2026 The Swarm Authors. All rights reserved.
+  SPDX-License-Identifier: Apache-2.0
+-->
+
+<script lang="ts">
+  import Trash2 from '@lucide/svelte/icons/trash-2'
+  import type { PostageStamp } from '@snaha/swarm-id'
+
+  import { Button } from '$lib/components/ui/button'
+  import { Dialog } from '$lib/components/ui/dialog'
+  import type { Account } from '$lib/types'
+
+  interface Props {
+    account: Account
+    drive: PostageStamp
+    onClose: () => void
+    onRemoved?: (message: string) => void
+  }
+
+  let { account, drive, onClose, onRemoved }: Props = $props()
+
+  function remove() {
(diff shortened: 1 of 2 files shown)


ui/src/lib/components/drive-remove-dialog.svelte
```diff
@@ -0,0 +1,42 @@
...
+    account.removeStamp(drive.batchID)
+    onRemoved?.('Drive removed')
+    onClose()
+  }
+</script>
+
+<Dialog onclose={onClose} title={drive.name || 'Drive'}>
+  <p class="text-sm">
+    This removes the drive from your account on all your devices. The storage itself stays on the
+    Swarm network — you can re-attach it later with its Batch ID.
```
agazso: This only works if the same account re-attaches a batch. If another account uses the batch the postage batch state is lost.

agazso (approved):

ui/src/lib/components/drive-remove-dialog.svelte
vojtechsimetka: Good point — copy now scoped to the same account in 6f9c42a2: "this account can re-attach it later with its Batch ID."

PR #398, opened 2026-07-03
fix(lib): migrate remaining rejectAfter races to withTimeout

### LINKED ISSUE #383: Leaked setTimeout handles: migrate remaining rejectAfter races to withTimeout
## Summary

Several timeouts are expressed as `Promise.race([work, rejectAfter(ms, msg)])`.
`rejectAfter` (`lib/src/utils/promise.ts`) never clears its `setTimeout`, so when
`work` wins the race the timer stays alive until it fires — holding its closure
and an event-loop reference. Harmless to correctness (rejecting an already-settled
race is a no-op; `Promise.race` attaches handlers, so no unhandled rejection), but
wasteful in hot paths.

#382 already fixed the worst offenders and introduced the right tool:
`withTimeout(work, ms, message)` in `lib/src/utils/promise.ts`, which races the
same way **and clears the timer in `.finally`**. This issue tracks migrating the
remaining sites.

## Sites that still leak

- `lib/src/utils/promise.ts` — `rejectAfter()` itself (root cause; never clears by
  construction). Re-exported publicly via `lib/src/index.ts`.
- `lib/src/sync/partition-intent.ts:~388` — `rejectAfter` in `readPartitionIntent`'s
  per-read `download()` race. Hottest path: looped over partitions × rivals ×
  epoch buckets, up to 2 attempts each.
- `lib/src/sync/sync-account.ts:~222` — `rejectAfter` for the 30s utilization-upload
  timeout.
- `lib/src/swarm-id-client.ts:~287` — `rejectAfter` for the proxy-init timeout
  (`firstConnectionInfoPromise`).
- `lib/src/sync/batch-write-coordinator.ts:~413` — an inline
  `new Promise((_, reject) => setTimeout(() => reject(...), ...))` raced at ~424 for
  the 45s partition-lease-acquire timeout; same leak, not via `rejectAfter`.

## Already correct (reference patterns)

- `lib/src/utils/promise.ts` — `withTimeout()` (clears in `.finally`).
- `lib/src/sync/sync-account.ts:~374` — manual `timer` + `clearTimeout(timer)` in a
  `finally` (the sync-result timeout).
- `lib/src/sync/{partition-state,partition-lease,device-roster}.ts` — migrated to
  `withTimeout` in #382.

## Suggested fix

1. Replace each `Promise.race([work, rejectAfter(ms, msg)])` with
   `withTimeout(work, ms, msg)`.
2. Convert the inline `batch-write-coordinator.ts` acquire timeout to
   `withTimeout(this.acquireWithSlotWait(), PARTITION_LEASE_ACQUIRE_TIMEOUT_MS, …)`.
3. `rejectAfter` is part of the public API (`lib/src/index.ts`), so don't hard-remove
   it. After the internal callers are migrated, either deprecate it (JSDoc
   `@deprecated`, point to `withTimeout`) or document the leak caveat. Drop the
   internal import where it's no longer used.

No behaviour change is expected — `withTimeout` rejects with the same `Error(message)`
on timeout, so existing `.catch`/message checks (e.g. `INTENT_TIMEOUT_MESSAGE`,
`device-roster`'s timeout-vs-404 discrimination) keep working.

## Verification

- `pnpm --filter @snaha/swarm-id typecheck && pnpm --filter @snaha/swarm-id lint`
- `pnpm --filter @snaha/swarm-id test` (intent / sync-account / coordinator suites)
- Optional: a `withTimeout` unit test asserting `vi.getTimerCount() === 0` after the
  work wins (mirrors the one added for `device-roster` in #382).

vojtechsimetka:
## Summary

`rejectAfter` never clears its `setTimeout`, so every race the work wins leaves a stray timer holding its closure until it fires. This migrates the remaining internal timeout races to `withTimeout` (introduced in #382, clears the timer once the race settles):

- `lib/src/sync/sync-account.ts` — 30s utilization-upload timeout
- `lib/src/swarm-id-client.ts` — proxy-init timeout (`firstConnectionInfoPromise`)
- `lib/src/sync/batch-write-coordinator.ts` — inline setTimeout-reject race for the 45s partition-lease-acquire timeout

`lib/src/sync/partition-intent.ts` (listed in the issue) was already migrated on main. `rejectAfter` stays exported as public API, now tagged `@deprecated` pointing to `withTimeout` with the uncancelled-timer caveat.

No behaviour change: `withTimeout` rejects with the same message (`TimeoutError extends Error`), and no call site keys on the plain-`Error` constructor or these message strings. Timer cleanup is already covered by the `withTimeout` fast-path test in `lib/src/utils/promise.test.ts`.

Closes #383

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -118,6 +118,11 @@ Authentication uses storage events: popup writes to localStorage → storage eve
 - **kebab-case** for all file and directory names
 - **Conventional commits**: `feat:`, `fix:`, `docs:`, etc.
 - **TypeScript execution**: Use `pnpx tsx` (not `npx ts-node`)
+- **Timeouts: use `withTimeout`** (`lib/src/utils/promise.ts`) — never `Promise.race` work against
+  `rejectAfter` or an inline `setTimeout` rejection: when the work wins, the losing timer stays
+  armed and leaks its handle. `withTimeout(work, ms, message)` clears the timer and rejects with
+  `TimeoutError`, so discriminate timeouts with `instanceof TimeoutError`, not by message.
+  `rejectAfter` is deprecated and kept only for the public API.
 
 ## Testing
 
--- a/lib/src/swarm-id-client.ts
+++ b/lib/src/swarm-id-client.ts
@@ -85,7 +85,7 @@ import { EthAddress, Identifier, PrivateKey, Topic } from "@ethersphere/bee-js"
 import { uint8ArrayToHex } from "./utils/hex"
 import { buildAuthUrl } from "./utils/url"
 import { isWebKit } from "./utils/browser"
-import { rejectAfter } from "./utils/promise"
+import { withTimeout } from "./utils/promise"
 
 const DEFAULT_TIMEOUT_MS = 30000
 const DEFAULT_INITIALIZATION_TIMEOUT_MS = 30000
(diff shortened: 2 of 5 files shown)


agazso (approved):
This looks like something that could be added to `AGENTS.md` as a rule

vojtechsimetka:
Added the rule to AGENTS.md in d9707273 (Code Style): use `withTimeout` instead of racing against `rejectAfter`/inline `setTimeout` rejections, and discriminate timeouts with `instanceof TimeoutError`.

PR #399, opened 2026-07-03
feat(ui): sign-in recovery, product account sync (publish + fold), existing-batch attach

### LINKED ISSUE #388: New UI: "I already have an account" should verify the account exists on the network
Today the "I already have an account" flow is **fully offline**: it derives the keypair from the recovery phrase and creates the account locally without ever contacting the network or checking whether the account exists (`ui/src/routes/account/import/+page.svelte` → `signIn()` → `startSignIn(generateName(), phrase)`; the access page then creates it with no fold). So a mistyped-but-valid phrase silently creates a new empty account, and no synced state (name/stamps/apps) is recovered.

**Desired (per Figma flow):**
- Enter recovery phrase (**frame 24**).
- Verify the account on the network — read its roster / device-state via `foldAccountFromSwarm` (the existing fold, currently only wired into `/dev` as `refreshAccountFromSwarm`).
- **Found → proceed** (**frame 23**) and recover the synced state (name, stamps, connected apps) rather than generating fresh defaults (ties into #387).
- **Not found → "account not found" screen** (**frame 27**).

**Considerations:**
- Distinguish "no account on the network" (→ frame 27) from "network unreachable" (→ error / retry, not a false not-found).
- Needs the Bee endpoint available in the product sign-in path (today the network settings live under `$lib/dev`).

Figma: https://www.figma.com/design/cb9GBgCZbRKmJe2CpgYCiQ/SwarmID-MVP?node-id=0-1&p=f&m=dev (frames 24 → 23 on success, 27 if not found)

Related: #387 (name recovery on sign-in).

### LINKED ISSUE #389: New UI: product app doesn't sync to/from Swarm (publish + fold are dev-only)
In the new UI, account/drive mutations persist locally but never reach Swarm in the shipping app: the sync publish hook (`setAccountsSyncHook(triggerSync)`) and the fold-back (`refreshAccountFromSwarm` → `foldAccountFromSwarm`) are wired **only** in `/dev`. So in the product UI:

- Renames, connected-app changes, and drive add/rename/resize/extend/remove don't publish to Swarm — peers never see them.
- Nothing folds peer state back, so another device's changes (and node truth like a batch's real `utilization` / `batchTTL`) never reconcile.

The LWW/tombstone machinery for this is already correct (per-field account clocks; the two-clock stamp merge added in #375); it's just not exercised outside `/dev`. Wiring product sign-in + a background publish/fold into the shipping app is the missing piece.

Related (facets of the same "product doesn't fold from Swarm" gap):
- #387 — account name is random per device / not recovered on sign-in.
- #388 — "I already have an account" doesn't verify/recover the account from the network.

### LINKED ISSUE #391: New UI: expiry missing for a drive added from an existing batch
A drive added by attaching an existing batch (not purchased through the widget) shows no expiry date / time-left label in the Storage tab.

The attach path is supposed to snapshot the TTL: `fetchExistingStamp` (`ui/src/lib/payment/bee.ts`) sets `batchTTL: batch.duration.toSeconds()`, and `describeDrive` (`ui/src/lib/drives.ts`) hides expiry only when `batchTTL` is `undefined` or ≤ 0. Candidate causes to investigate:

- the stamp record predates the `batchTTL`/`updatedAt` fields (e.g. created via `/dev` or synced from an older record), so the snapshot was never taken
- the node reported `batchTTL: -1` (it can't compute the TTL, e.g. chain state unavailable), which bee-js passes through as a negative duration — the UI then treats the drive as expired/unknown rather than fetching a usable value

Reported in the review of #375 (with screenshot): https://github.com/snaha/swarm-id/pull/375#issuecomment-4868281602

agazso:
Related new-UI features that together make an account (and its drives) usable across devices and against real/independent postage batches, and wire product ↔ Swarm sync.

## Sign-in verifies & recovers the account (Closes #388)

"I already have an account" was fully offline — it derived the keypair and created a fresh empty account without ever touching the network, so a mistyped-but-valid phrase silently created a new account and no synced state was recovered.

- Sign-in now folds the account from Swarm first: **found** → recover its name, stamps and connected apps and proceed; **not found** → "double-check your phrase"; a `bee.isConnected()` probe distinguishes *no account* from *network unreachable*.

## Product ↔ Swarm account sync (Closes #389)

Publish and fold were both wired only in `/dev`, so the shipping app never synced.

- **Publish** — `setAccountsSyncHook(triggerSync)` is installed app-wide (root layout) and fires after account creation, so renames / connected-app changes / drive add·rename·resize·extend·remove now publish (debounced).
- **Fold-back** — the active account is folded from Swarm on **page load / account switch** and on a periodic **~3 min** tick (`refreshAccountFromSwarm`), so peer edits and node truth reconcile. Folds apply with `skipSync`, so they never re-publish.
- **Single fetcher across tabs** — a new lib primitive `runCoalescedAcrossTabs` (exclusive Web Lock + shared localStorage cooldown timestamp) makes only one tab fetch per window; the winner's persisted account document propagates to the other tabs via the existing accounts `storage` listener (unit-tested).

Follow-up (not blocking): the fold's read amplification/slowness against a remote gateway is tracked in #400.

## Attach existing postage batches, with network settings (Closes #391)

- **Network settings dialog** (Bee node URL + Gnosis RPC), ported from the legacy UI with new-UI components, wired into the settings menu and reachable while signed in.
- **Optional signer-key field** on "use existing batch" so a batch owned by an external key can be attached; both fields validate through the lib's Zod schemas.
- Existing-batch details are read from the **PostageStamp contract on-chain** (`fetchExistingBatchFromChain`) instead of the Bee node's `/stamps` (a public gateway doesn't expose it and never saw an independently-bought batch). This also snapshots the batch **TTL from the contract**, so an existing-batch drive shows its expiry (**#391**).
- A single **stamped test chunk** (`verifyBatchStampable`) confirms the signer can actually stamp for the batch before attaching; only a definitive node rejection blocks it, so a slow pushsync never turns a valid batch away.
- Restores the `/dev` **"Import batch by ID"** tool, sharing the same on-chain lookup.

### Note on utilization / bucket state of an attached batch

An attached batch's record is created with `utilization: 0` (the contract carries no per-bucket usage). What actually governs stamping correctness — the per-bucket counters — is handled elsewhere and is **not** regressed by this PR:

- **Same account, same browser** (e.g. remove a drive, then re-add the same batch): the bucket counters live in the local IndexedDB util store, which remove only tombstones the account record — it never purges. So stamping stays correct, and the displayed % (which momentarily shows 0 on re-add) self-corrects on the next sync.
- **Same account, multi-device**: the real upload path recovers the cross-device counter from the on-Swarm partition-state feed on the partition-lease acquire that gates every stamped write — unchanged by this PR.
- **Known limitation**: a batch owned by an **external account** (pasted external signer) can't recover its exact bucket state on attach — the on-Swarm state is addressed by the *publishing* account's encryption key, which we don't have. So `verifyBatchStampable` probes from a zeroed bucket state (a heavily-used external batch could get a spurious rejection) and the displayed utilization starts at 0. Left as a follow-up.

## Verification

`pnpm check:all` passes (incl. the `runCoalescedAcrossTabs` unit test). Exercised end-to-end against the local bee cluster: sign-in not-found/unreachable paths; app-wide publish + on-load/periodic fold (single-fetcher timestamp advancing once per window); add-drive with an existing batch (correct signer attaches, wrong signer is rejected, details/expiry read via `eth_call` with no `/stamps` request); the `/dev` import tool; and the network-settings dialog (validation, persistence, reset).

### DIFF
--- a/lib/src/index.ts
+++ b/lib/src/index.ts
@@ -28,6 +28,9 @@ export {
 // Hex address utility
 export { hexAddress } from "./utils/hex"
 
+// Cross-tab coalescing: run a task at most once per window across same-origin tabs
+export { runCoalescedAcrossTabs } from "./utils/coalesced-task"
+
 // Stamp worker pool for parallel ECDSA signing
 export { StampWorkerPool } from "./proxy/stamp-worker-pool"
 
--- a/lib/src/index.ts
+++ b/lib/src/index.ts
@@ -231,6 +231,7 @@ export {
   ensureInRoster,
   ROSTER_TOPIC_PREFIX,
   foldAccountFromSwarm,
+  foldedToAccountData,
 } from "./sync"
 
 // State sync types
@@ -457,6 +458,7 @@ export { downloadDataWithChunkAPI } from "./proxy/download-data"
 export { downloadEncryptedSOC } from "./proxy/download-data"
 export {
   uploadSOC,
+  SocUploadError,
   type UploadTarget,
   type UploadSOCOptions,
 } from "./proxy/upload"
@@ -480,7 +482,7 @@ export type { ActEntry } from "./proxy/act"
 export { SWARM_SECRET_PREFIX, STORAGE_CHALLENGE_KEY } from "./types"
 
 // URL building utilities
-export { buildAuthUrl } from "./utils/url"
+export { buildAuthUrl, isHttpUrl } from "./utils/url"
 
 // Browser detection utilities
 export { isWebKit } from "./utils/browser"
--- a/lib/src/proxy/upload.ts
+++ b/lib/src/proxy/upload.ts
@@ -43,6 +43,26 @@ import { uint8ArrayToHex } from "../utils/hex"
 import { normalizeUrl } from "../utils/url"
 
 // ============================================================================
+// Errors
+// ============================================================================
+
+/**
+ * Thrown when a SOC upload gets a non-OK HTTP response. Carries the numeric
+ * `status` so callers can distinguish a definitive stamp rejection (400) from a
+ * transient / not-yet-synced condition (404 "batch not found", 422 "not usable
+ * yet") without regex-matching the message — see `verifyBatchStampable`.
+ */
+export class SocUploadError extends Error {
+  readonly status: number
+
+  constructor(status: number, statusText: string, body: string) {
+    super(`SOC upload failed: ${status} ${statusText} - ${body}`)
+    this.name = "SocUploadError"
+    this.status = status
+  }
+}
+
+// ============================================================================
 // Types
 // ============================================================================
 
@@ -910,8 +930,10 @@ export async function uploadSOC(
 
     if (!response.ok) {
       const errorText = await response.text()
-      throw new Error(
-        `Subsidised SOC upload failed: ${response.status} ${response.statusText} - ${errorText}`,
+      throw new SocUploadError(
+        response.status,
+        `${response.statusText} (subsidised)`,
+        errorText,
       )
     }
 
@@ -955,9 +977,7 @@ export async function uploadSOC(
 
     if (!response.ok) {
       const errorText = await response.text()
-      throw new Error(
-        `SOC upload failed: ${response.status} ${response.statusText} - ${errorText}`,
-      )
+      throw new SocUploadError(response.status, response.statusText, errorText)
     }
 
     return {
--- a/lib/src/schemas.ts
+++ b/lib/src/schemas.ts
@@ -15,6 +15,7 @@ import {
   BatchId as BeeBatchId,
   PrivateKey as BeePrivateKey,
 } from "@ethersphere/bee-js"
+import { isHttpUrl } from "./utils/url"
 
 // ============================================================================
 // Network Settings Constants
@@ -448,8 +449,11 @@ export type PartitionIntentPayload = z.infer<
  * Stores user-configurable network endpoints (Bee node and Gnosis RPC)
  */
 export const NetworkSettingsSchemaV1 = z.object({
-  beeNodeUrl: z.string().url(),
-  gnosisRpcUrl: z.string().url(),
+  // `.refine(isHttpUrl)` (not `.url()`) so a scheme-less paste like `localhost:1633`
+  // — which `new URL`/zod `.url()` accept as protocol `localhost:` — can't persist
+  // as a broken endpoint. Same validator the settings dialog uses.
+  beeNodeUrl: z.string().refine(isHttpUrl, "Must be an http(s) URL"),
+  gnosisRpcUrl: z.string().refine(isHttpUrl, "Must be an http(s) URL"),
 })
 
 export type NetworkSettings = z.infer<typeof NetworkSettingsSchemaV1>
--- a/lib/src/sync/fold-account-from-swarm.ts
+++ b/lib/src/sync/fold-account-from-swarm.ts
@@ -8,7 +8,7 @@
  * / refresh / the proxy.
  */
 
-import { Bee, PrivateKey } from "@ethersphere/bee-js"
+import { Bee, EthAddress, PrivateKey } from "@ethersphere/bee-js"
 import { deriveSwarmEncryptionKey, deriveSecret } from "../utils/key-derivation"
 import { readRoster } from "./device-roster"
 import {
@@ -17,11 +17,48 @@ import {
   type DeviceStateSnapshot,
   type FoldedAccount,
 } from "./device-state"
-import type { Device } from "../schemas"
+import type { AccountData, Device } from "../schemas"
 
 export interface FoldAccountResult {
   account: FoldedAccount
   devices: Device[]
+}
+
+/**
+ * Project a folded account onto the local `AccountData` record — the single
+ * place the field list lives so a new `AccountData`/`FoldedAccount` field can't
+ * be silently dropped by a hand-written copy on the sign-in path. `id` and
+ * `derivationKey` aren't in the fold (they come from the recovering wallet), so
+ * the caller supplies them.
+ *
+ * `foldAccountFromSwarm` shallow-FREEZES its result arrays so coalesced callers
+ * can't corrupt each other's shared view; a persisted `AccountData` outlives that
+ * window and is mutated in place downstream, so the arrays are COPIED here.
+ */
+export function foldedToAccountData(opts: {
+  id: EthAddress
+  derivationKey: string
+  account: FoldedAccount
+  lastModified?: number
+}): AccountData {
+  const { id, derivationKey, account } = opts
+  return {
+    id,
+    name: account.accountName,
+    publicKey: account.publicKey,
+    createdAt: account.createdAt,
+    derivationKey,
+    defaultPostageStampBatchID: account.defaultPostageStampBatchID,
+    settings: account.settings,
+    accountNameAt: account.accountNameAt,
+    defaultStampAt: account.defaultStampAt,
+    settingsAt: account.settingsAt,
+    lastModified: opts.lastModified ?? Date.now(),
+    devices: [...account.devices],
+    connectedApps: [...account.connectedApps],
+    postageStamps: [...account.postageStamps],
+    partitionCount: account.partitionCount,
+  }
 }
 
 /**
--- a/lib/src/sync/index.ts
+++ b/lib/src/sync/index.ts
@@ -57,7 +57,10 @@ export {
   ensureInRoster,
   ROSTER_TOPIC_PREFIX,
 } from "./device-roster"
-export { foldAccountFromSwarm } from "./fold-account-from-swarm"
+export {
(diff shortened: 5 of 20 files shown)


vojtechsimetka:
I went through this with a focus on code quality and duplication. Overall a nice set of features — sign-in finally touching the network is a big UX win and `runCoalescedAcrossTabs` with tests is a good addition. A few things I think we should fix before merging though, mostly in the attach path.

## Should be fixed before merge

1. `ui/src/lib/payment/contract.ts:61` — `amount: batch.normalisedBalance` is the wrong unit. `normalisedBalance` is the cumulative outpayment watermark (remaining balance + global `currentTotalOutPayment` since genesis), but everything downstream treats `stamp.amount` as the per-chunk balance: `dilutedStamp` computes the keep-lifespan top-up from it and the resize dialog actually pays it via `topUpStamp`. So resizing an attached drive quotes — and when the node owns the batch, pays — a top-up inflated by multiples of a normal purchase, and the wrong amount then syncs to all devices. The comment "record metadata only — the stamper doesn't read it" doesn't hold, dilute/extend do read it. We should store `normalisedBalance − currentTotalOutPayment` (both are already in `OnChainBatchState`).

2. `ui/src/routes/account/import/+page.svelte:83` — an account that never had a stamp can't sign in on a new device anymore, and we blame the phrase. Publishing requires a default stamp (`getAccountStateSnapshot` bails without one), so such an account has no roster on Swarm, the fold returns `undefined`, and a user with a *correct* phrase gets "double-check your secret recovery phrase". Same device is fine since the local store is checked first. I think we need either an escape hatch ("no account found on the network — it may never have synced, you can still set it up fresh") or at least copy that doesn't accuse the phrase.

3. `ui/src/lib/payment/bee.ts:38` — `NODE_REJECTED` matches *any* 4xx, but Bee also returns 404 "batch not found" / 422 "batch not usable yet" for a batch the node simply hasn't synced or warmed up yet — which is exactly the scenario this feature exists for. A freshly bought, perfectly valid batch gets a definitive "Check the signer key". Matching on the error message string from another package is also fragile: any rewording in `uploadSOC` silently turns definitive rejections into "inconclusive" (attach proceeds). The lib already has typed errors (`TimeoutError`, `SnapshotDataUnavailableError`, …) — `uploadSOC` should throw one carrying the HTTP status, and only the stamp-validation 400 should count as definitive.

4. `ui/src/lib/payment/bee.ts:55` — every `verifyBatchStampable` call burns a postage slot that nothing records. The throwaway `Stamper.fromBlank` state is discarded, and neither the utilization store nor the partition-state feed ever learns about the consumed (bucket, index 0). The first real chunk that hashes into that bucket re-issues the same slot, which storer nodes reject as overissuance on an immutable batch. 1/65536 per chunk, but pretty much guaranteed over a drive's lifetime, and each re-validation poisons another bucket. Could we seed the utilization cache with the consumed slot, or probe a reserved slot the way the lock-SOC does?

5. `ui/src/routes/account/import/+page.svelte:96` — this 15-field `FoldedAccount` → `AccountData` copy duplicates the projection that already lives in the lib as the private `foldedToSnapshot` (`lib/src/sync/restore-account.ts:92`), including the easy-to-miss "the fold freezes its arrays" copying. And since the fold schema fields have zod defaults, a future field won't type-error here — it just gets silently dropped for accounts imported on a new device. The lib should export a `foldedToAccountData` next to `foldAccountFromSwarm` so the field list lives in one place.

## Smaller things

6. `ui/src/lib/components/network-settings-dialog.svelte:23` — `isUrl` duplicates what the store's schema already validates, and both accept `localhost:1633` (`new URL` parses it with protocol `localhost:`, and zod's `.url()` takes it too — I checked), so the comment about the schema rejecting malformed values on next load doesn't hold: a scheme-less paste gets saved and persists as a broken Bee URL. One shared validator that also requires `http:`/`https:` fixes both.

7. `ui/src/lib/dev/account-refresh.ts:37` — the module-level `inFlight` silently drops the forced fold on account switch: switch A→B while A's fold is still running (your own comment says overlapping triggers are the norm) and B shows stale state until the next 3-minute tick. The lib's `inFlightFolds` is already keyed per account — key this one too, or re-run the pending id in `finally`.

8. `LOCAL_POSTAGE_STAMP_CONTRACT_ADDRESS` is defined in both `ui/src/lib/payment/contract.ts:17` and `ui/src/routes/dev/+page.svelte:590` (and a third time in `package.json`). Duplicates, should live in one place — export it from `contract.ts`, or let the lib resolver own the local default.

## Nits

- The layout `$effect` always folds with `force: true`, so N restored tabs run N sequential full folds (the Web Lock serialises, the cooldown is bypassed) — and every new-device sign-in folds twice back-to-back (import page directly, then the forced fold after finalize). A short grace window under `force`, or stamping the cooldown after the import-page fold, would fix both.
- `ui/src/lib/components/drive-add-dialog.svelte:283` — the pasted-signer check is done in both `proceed()` and `attachExisting()`, with `seed as Uint8Array` hiding the invariant; one `$derived` for the trimmed key plus an explicit guard would remove the cast. Also, with a valid batch ID and a malformed signer key the button just disables with no hint — `infoText` only looks at `batchIdValid`.
- `ui/src/routes/+layout.svelte:41` — the `setTimeout(0)` + `clearTimeout` dance is just `untrack(() => void foldCurrentAccount(true))`.
- `ui/src/lib/components/network-settings-dialog.svelte:64` — `X.trim().length > 0 && !XValid` is repeated four times; two `$derived`s would do.
- `ui/src/lib/payment/contract.ts:24` — the `ExistingBatchLookup` union doesn't say more than `NewStamp | undefined`; keep it only if an `rpc-error` arm is actually planned.
- `ui/src/routes/account/import/+page.svelte:76` — `bee.isConnected()` is a serial extra round-trip whose answer only matters in the not-found branch; could run concurrently with the fold.

Things I checked and deliberately didn't flag: re-import overwriting local state (guarded by the early `accountsStore.get` return), `utilization: 0` (acknowledged in the description, and real bucket state recovers via the partition-lease path), stamping the cooldown even on failure (intentional, there's a test for it), and the sync engine living under `$lib/dev` (deferred to #389 in-code, fine by me).

agazso:
Thanks for the thorough pass — all addressed in 735ff23e. Rundown:

**Should-fix**

1. **`contract.ts` amount unit** — now stores `normalisedBalance − currentTotalOutPayment` (both from `OnChainBatchState`), i.e. the remaining per-chunk balance that dilute/extend read. Comment corrected.
2. **Sign-in on a never-synced account** — added an escape hatch: a distinct "No account found" screen (copy no longer accuses the phrase) with **"Set up on this device"**, which proceeds to the access step and builds a fresh account under the same phrase. "Try a different phrase" is the other action.
3. **`bee.ts` 4xx over-matching** — `uploadSOC` now throws a typed `SocUploadError` carrying the HTTP `status` (kills the cross-package message regex). `verifyBatchStampable` treats **only a 400** as definitive; 404/422/timeout/anything else is inconclusive, so a freshly-bought / not-yet-synced batch still attaches. (Block-number freshness disambiguation left out for now — noted as a follow-up.)
4. **`verifyBatchStampable` slot burn** — not an issue in practice: the batches we use are **mutable**, so overstamping a reserved slot is legal (no overissuance). Left the probe as-is.
5. **Duplicated `FoldedAccount → AccountData` projection** — the lib now exports `foldedToAccountData` next to `foldAccountFromSwarm` (the field list + frozen-array copy live in one place); the import page uses it.

**Smaller**

6. **URL validation** — added `isHttpUrl` to the lib, now shared by `NetworkSettingsSchemaV1` (`.refine(isHttpUrl)`, not `.url()`) and the dialog, so a scheme-less `localhost:1633` fails both and can't persist.
7. **`account-refresh` inFlight** — now a per-account `Set`, so an A→B switch mid-fold isn't dropped.
8. **`LOCAL_POSTAGE_STAMP_CONTRACT_ADDRESS`** — exported from `contract.ts` and imported by the dev page. (The `package.json` env-var copy is a separate build-time mechanism, left as-is.)

**Nits**

- Double-fold: `runCoalescedAcrossTabs` gained a `graceMs` window honoured **even under `force`**, so N restored tabs collapse to one fold; the import-page fold stamps the cooldown (`noteAccountFolded`) so the post-finalize forced fold skips within the window.
- `drive-add-dialog`: single `$derived` for the pasted signer key removes the `seed as Uint8Array` cast (explicit guard instead); `infoText` now hints when the batch ID is valid but the signer key is malformed.
- `+layout`: `setTimeout(0)` + `clearTimeout` → `untrack(() => void foldCurrentAccount(true))`.
- `network-settings-dialog`: two `$derived` error flags replace the four repeated conditions.
- `ExistingBatchLookup` union → `NewStamp | undefined` (no `rpc-error` arm planned).
- `import`: `bee.isConnected()` now runs concurrently with the fold; only awaited in the not-found branch.

`pnpm check:all` is green.

vojtechsimetka (approved):

Issue #412, opened 2026-07-06
ui: decrypted seed entropy cached indefinitely — later sensitive actions skip re-auth


vojtechsimetka:
## Problem
After one unlock, `entropy` (`ui/src/lib/components/home-account.svelte:89`) holds the raw BIP-39 entropy for the component's lifetime and `openUnlock()` short-circuits when it's set. So after revealing the phrase once, **export backup**, **change unlock method**, and **delete account** all proceed with no confirmation ceremony. `drive-add-dialog.svelte:74` caches the same material in a non-reactive local.

## Fix
Scope the cached entropy per-target or add an idle TTL that clears it, and zero the `Uint8Array` on clear (both here and in drive-add-dialog).
---
_Filed from an internal code review of `lib/` and `ui/`. Findings are from a focused read and should be validated before acting; the ACT crypto items in particular should be checked against upstream Bee's construction._

PR #429, opened 2026-07-06
feat(ui): connect screen for devices with existing accounts

vojtechsimetka:
Implements the connect-popup screen for the case where accounts already exist on the device, per the [Figma design](https://www.figma.com/design/RyprPgEZIeCoaqjTcKXGUo/SwarmID-MVP--Copy-?node-id=159-14450&m=dev) (frame 159-14450), replacing the previous tabs-based layout.

**Scope note:** the destructive account actions (sign out / remove / delete) are still being settled by the team — incl. whether deletion must dilute the batch — so **Remove an account is stubbed** with `notImplemented` here; the remove mode + confirmation land with that follow-up.

## What's included

- **Account list** — one bordered row per device account (identicon, name, truncated address), ordered by most recent connection to the requesting app; never-connected accounts keep their stored order. Selection behavior is unchanged: a still-valid session reconnects instantly, otherwise the unlock dialog (password / passkey / ETH wallet) opens.
- **"Signed out" badge** — shown on accounts that were previously connected to this app but whose session lapsed (expired or revoked), signalling that reconnecting will ask for an unlock. Backed by a new `hasReusableConnection` helper in `connect-handshake.ts` (shared with `reuseConnection`).
- **Remove an account** — button per the design, stubbed `notImplemented` (see scope note).
- **Add an account** — swaps the list for the create/import choice (the same view a device with no accounts gets), with a back button to return to the list.
- Ghost-button label updated to the design's **"What is Swarm ID?"** (was "WTF is Swarm ID?").

The no-request and no-accounts states, the unlock ceremony, and `/connect/done` are unchanged.

## Testing

- `pnpm check:all` green across packages.
- Exercised in-browser against the dev server with real accounts created through the setup flow: instant reconnect (valid session), unlock with wrong + correct password, add-account view and back, badge + ordering, Remove an account alerts not-implemented; light and dark themes; no console errors.

## Screens

**Connect — accounts on device (badge on lapsed session)**

Connect with accounts

**Add an account**

Add an account

**Unlock to connect**

Unlock dialog

**Dark mode**

Dark mode

### DIFF
--- a/ui/src/lib/connect-handshake.ts
+++ b/ui/src/lib/connect-handshake.ts
@@ -86,19 +86,29 @@ export async function completeConnect(
   sendSecretToOpener(account, request, appSecret)
 }
 
+/** The still-valid prior connection to this app, if the account has one. */
+function findReusableConnection(account: Account, appOrigin: string): ConnectedApp | undefined {
+  return account.connectedApps.find(
+    (app) =>
+      app.appUrl === appOrigin &&
+      app.appSecret !== undefined &&
+      app.revokedAt === undefined &&
+      (app.connectedUntil ?? 0) > Date.now(),
+  )
+}
+
+/** Whether connecting can skip the unlock ceremony (see `reuseConnection`). */
+export function hasReusableConnection(account: Account, appOrigin: string): boolean {
+  return findReusableConnection(account, appOrigin) !== undefined
+}
+
 /**
  * Reconnect with the secret of a still-valid prior connection, skipping the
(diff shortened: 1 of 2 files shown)


agazso (approved):
This works for now, but it does not ask for authentication when connecting an app: #412 

Maybe after fixing that issue, this needs to be fixed again.

vojtechsimetka:
For now I actually put the delete and sign out as a not implemented methods. These should all be revisited in a bigger PR

PR #452, opened 2026-07-07
fix(lib)!: remove unsafe utilization-cache LRU eviction

### LINKED ISSUE #419: lib(storage): utilization-cache LRU eviction can delete live counters and cause overstamping
## Problem
`evictOldEntries` (`lib/src/storage/utilization-store.ts:392`) LRU-deletes chunk rows (`maxChunks` default ~640). For a still-active batch, `loadUtilizationState` then seeds the missing range as zeroed/dirty, so the stamper restarts that bucket at slot 0 and **overstamps existing user data** (Bee evicts the older chunk). Partition-state resume mitigates this for partitioned accounts, but legacy single-device accounts have no on-Swarm counter to recover from. Also `const db = (cache as never)["db"]` (line 401) is a private-field escape hatch.

## Fix
Exclude currently active/attached batches from eviction, or evict per-batch atomically with a "counter lost, must resync" marker. Expose `db` via a proper method.
---
_Filed from an internal code review of `lib/` and `ui/`. Findings are from a focused read and should be validated before acting; the ACT crypto items in particular should be checked against upstream Bee's construction._

agazso:
## Problem

`evictOldEntries` LRU-deleted chunk rows of the utilization store. For a still-active batch, `loadUtilizationState` then re-seeds the missing range as zeroed/dirty, so the stamper restarts those buckets at slot 0 and **overstamps existing user data** (Bee evicts the older chunk at a colliding stamp index). It also reached into the store's private `db` field via `(cache as never)["db"]`.

## Why removal instead of hardening

- **Zero callers**: `evictOldEntries` was exported but never wired up anywhere in the monorepo (shipped with #111).
- **No safe eviction exists**: for legacy single-device accounts the local counter is *authoritative* — there is no on-Swarm state to resync from, so both suggested mitigations (active-batch allowlist, "must resync" marker) still end in counter loss or a permanent error.
- **No growth problem**: counters are ≤256KB per batch; `clearBatch` remains for explicit removal.

The unused `lastAccess` index is dropped from the `onupgradeneeded` handler — no `DB_VERSION` bump needed (fresh DBs won’t create it; existing v3 DBs keep a harmless orphan index nothing queries). Deliberately not bumping to v4: the upgrade handler drops both stores, which would wipe authoritative local counters.

**BREAKING CHANGE**: `evictOldEntries` and `CacheEvictionPolicy` are no longer exported from `@snaha/swarm-id`.

Closes #419

### DIFF
--- a/lib/src/storage/utilization-store.ts
+++ b/lib/src/storage/utilization-store.ts
@@ -153,7 +153,8 @@ export class UtilizationStoreDB {
         // Index for querying by batchId
         chunksStore.createIndex("batchId", "batchId", { unique: false })
 
-        // Index for eviction by lastAccess
+        // Unused since LRU eviction was removed (#419); kept to avoid a
+        // schema version bump just to drop an index.
         chunksStore.createIndex("lastAccess", "lastAccess", {
           unique: false,
         })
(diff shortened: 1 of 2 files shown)


lib/src/storage/utilization-store.ts
```diff
@@ -153,7 +153,8 @@ export class UtilizationStoreDB {
         // Index for querying by batchId
         chunksStore.createIndex("batchId", "batchId", { unique: false })
 
-        // Index for eviction by lastAccess
+        // Unused since LRU eviction was removed (#419); kept to avoid a
+        // schema version bump just to drop an index.
```
vojtechsimetka: maybe we can still drop the index now without any issues?

vojtechsimetka (approved):

lib/src/storage/utilization-store.ts
agazso: Dropped in 4aaf85bb0 — no version bump needed; existing v3 DBs just keep a harmless orphan index. Not bumping to v4 since the upgrade handler drops both stores and would wipe local counters.

Issue #338, opened 2026-06-04
Account-state restore/refresh coverage: agent restore + refresh on load/account-switch


agazso:
Two gaps in how account state reaches / stays current on a device. Both are the **read/pull side** of multi-device sync (getting published state onto a device and keeping it fresh), so they're grouped here. Found during the #335 work.

## 1. Agent restore path

Agent (seed-phrase) accounts create **fresh local state** on a new device instead of restoring from Swarm. The restore path (`restoreAccountFromSwarm` → `restoreAccountToStores`) is wired only into **passkey/ethereum** sign-in (`swarm-ui/src/routes/(app)/(create)/signin/passkey`, `signin/ethereum`, `swarm-ui/src/lib/utils/restore-account.ts`); agent creation (`(create)/agent/new`) just creates a new account locally. So a 2nd agent device never pulls the account's published identities / connected apps / stamps / device list.

**Fix:** give agent accounts a restore-from-Swarm path on (re-)entry — mirror the passkey/ethereum sign-in restore, or add an explicit "agent sign-in" flow.

## 2. Broaden refresh triggers

`refreshAccountFromSwarm` is currently invoked only from the Devices view (moved to `/dev` → Devices) on mount. An already-known device therefore goes **stale** — changes made on a peer (new identity, connected app, stamp, device) don't appear until the user happens to open that view.

**Fix:** call `refreshAccountFromSwarm` on app load and on account switch (e.g. from the `(app)` root layout or an account-switch hook), not just the Devices view. Consider applying the full snapshot there too (the refresh already merges all sections into the stores via `applyRefreshed*`).

## Relationship

- Part of #217 (multi-device umbrella).
- Complements #209 — that tracks the **write/publish-trigger** family ("sync should be triggered when …"); this issue is the **read/pull-trigger** side.
- Independent of #336 (proxy-publish / coordinator) and #337 (merge dedup).
- Context: introduced alongside #335.

## Acceptance

- A fresh agent device restores the account's state from Swarm (identities, apps, stamps, devices), not a blank account.
- An already-signed-in device picks up peer changes without visiting the Devices view (refresh on load / account switch).
- `pnpm check:all` green.

agazso:
Both asks landed a while ago, and the last gap is fixed in #456 — closing with the evidence trail:

[...]

PR #456, opened 2026-07-07
fix: converge dApp-only devices on peer account changes

agazso:
Part of #338 (read/pull triggers) — closes its last remaining gap.

## Problem

A device where the user only interacts through a dApp (the identity UI never opened after the first connect) never converged on peer changes: a rename, default-stamp change, or drive edit made on another device stayed invisible **forever**. Two root causes:

1. The fold engine only ever folded the *session-current* account. The root layout (and its 3-min fold tick) does mount inside the `/proxy` iframe, but the account connected to the dApp need not be the current one — and the fold never covered the rest.
2. Even when a fold ran inside the proxy iframe, the running `SwarmIdProxy` never saw it: `VersionedStorageManager.subscribe` relied solely on the browser `storage` event, which only fires in **other** windows. The proxy caches its auth/stamper and re-derives only via that subscription, so a same-window fold write was invisible to it.

## Changes

- **lib**: `VersionedStorageManager.save()`/`clear()` now broadcast a window-scoped `CustomEvent`; other manager instances for the same key in the same window relay it to their subscribers (the writing instance skips itself). Unit-tested (TDD).
- **ui**: the periodic fold tick now folds **every** local account (per-account cross-tab cooldown unchanged), and `startFoldInterval` runs one immediate cooldown-respecting pass so a freshly loaded page converges right away.

## Testing

- `pnpm check:all` green (941 lib tests incl. 6 new versioned-storage tests).
- E2E against the local bee cluster with a scripted "device A" peer (`scripts/publish-device-a.ts`, untracked) and Playwright as device B:
  1. **Fresh-device restore**: seed-phrase sign-in restored the published name, stamp (set as default), and peer device — not a blank account.
  2. **dApp-only convergence**: connected the demo dApp, closed all identity-UI tabs, renamed the account from the peer script → the proxy iframe's fold pulled the rename, the same-window notification woke the proxy, and the demo sidebar re-rendered to the new name live — no reload, no identity-UI visit.

## Pre-existing finding (not addressed here)

On a small dev cluster, bee answers *absent-chunk* retrievals with `500 no peer found` instead of `404`. `ensureInRoster` treats that as an inconclusive roster read and skips the bootstrap append — so a brand-new account can never write roster slot 0 against a local cluster (restore/fold then fail with `RosterScanInconclusiveError`). Worked around in the e2e by force-writing slot 0; worth its own issue.

### DIFF
--- a/lib/src/utils/versioned-storage.test.ts
+++ b/lib/src/utils/versioned-storage.test.ts
@@ -1,0 +1,118 @@
+// Copyright 2026 The Swarm Authors. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest"
+import { z } from "zod"
+import {
+  VersionedStorageManager,
+  MemoryStorageAdapter,
+  createZodParser,
+  type StorageAdapter,
+} from "./versioned-storage"
+
+interface Item {
+  name: string
+}
+
+const ItemsSchema = z.array(z.object({ name: z.string() }))
+
+function makeManager(
+  storage: StorageAdapter,
+  key = "test-key",
+): VersionedStorageManager<Item> {
+  return new VersionedStorageManager<Item>({
+    key,
+    currentVersion: 1,
+    storage,
+    parsers: { 1: createZodParser(ItemsSchema) },
+  })
+}
+
+describe("VersionedStorageManager — same-window change notification", () => {
+  // The browser's `storage` event only fires in OTHER windows. Two manager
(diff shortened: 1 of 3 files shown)


agazso:
The pre-existing finding noted in the description (dev-cluster absent-chunk 500s deadlocking the fresh-account roster bootstrap) is now tracked in #457 — root-caused to the roster probe reading `/chunks` instead of `/soc`, which breaks #438's clean-404 premise.

vojtechsimetka (approved):

PR #484, opened 2026-07-13
fix(ui): dialog focus management and tabs keyboard contract

### LINKED ISSUE #426: ui: UI primitives missing accessibility — Dialog focus trap and Tabs keyboard contract
## Problem
The hand-written primitives announce roles they don't implement:
- **Dialog** (`ui/src/lib/components/ui/dialog/dialog.svelte:35`) — `role="dialog" aria-modal="true"` on a plain `div` with no initial focus, no focus trap, and no focus restore. Tab walks the obscured page behind the overlay; keyboard users can type a password into a background control. Inherited by every unlock/password dialog.
- **Tabs** (`ui/src/lib/components/ui/tabs/tabs.svelte:29`) — `role="tablist"/"tab"` with no roving tabindex, no Left/Right arrow navigation, no `aria-controls`/tabpanel linkage.

## Fix
Render a native `<dialog>` + `showModal()` (free focus trap, Escape, top-layer) or add a manual trap + `autofocus` + focus restore. For Tabs add roving tabindex + arrow-key handler, or drop the tab roles where it's really a segmented control.
---
_Filed from an internal code review of `lib/` and `ui/`. Findings are from a focused read and should be validated before acting; the ACT crypto items in particular should be checked against upstream Bee's construction._

vojtechsimetka:
Implements the accessibility contracts the hand-written primitives were announcing but not providing (per WAI-ARIA APG, no new dependencies).

**Dialog** (`ui/src/lib/components/ui/dialog/dialog.svelte`)
- On open, focus moves into the dialog: a consumer-marked `[data-autofocus]` element, else the first focusable control, else the panel itself
- Tab/Shift+Tab wrap within the dialog, and pull focus back inside if it ever lands outside
- On close, focus returns to the opener if it still exists — the restore chains correctly through pending-state dialog swaps (unlock → pending → back)
- The unlock dialogs mark their password input `data-autofocus`, so keyboard users can type the password immediately

**Tabs** (`ui/src/lib/components/ui/tabs/tabs.svelte`)
- Roving tabindex: only the selected tab is a page Tab stop
- ArrowLeft/ArrowRight (with wrap-around) and Home/End move focus and selection together (automatic activation)

**Tests** (Playwright, `ui/tests/`)
- `dialog-a11y.test.ts` — initial focus, Tab wrap in both directions, Escape, `data-autofocus`, focus restore to the opener
- `tabs-a11y.test.ts` — arrow/Home/End navigation with wrap, automatic activation, single-Tab-stop roving

All dialog and tabs consumers verified manually in the browser (network settings, unlock, set-method with tabs inside a dialog, delete account), plus the existing Playwright suite passes.

Out of scope: `aria-controls`/`role="tabpanel"` linkage — panels are rendered by consumers outside the primitive.

Closes #426

### DIFF
--- a/ui/src/lib/components/home-account.svelte
+++ b/ui/src/lib/components/home-account.svelte
@@ -635,6 +635,7 @@
           bind:value={password}
           placeholder="Account password"
           autocomplete="current-password"
+          data-autofocus
           onkeydown={(event: KeyboardEvent) => event.key === 'Enter' && confirmUnlock()}
         />
       {/if}
--- a/ui/src/lib/components/ui/dialog/dialog.svelte
+++ b/ui/src/lib/components/ui/dialog/dialog.svelte
@@ -3,8 +3,14 @@
   SPDX-License-Identifier: Apache-2.0
 -->
 
+<!--
+  Modal dialog implementing the WAI-ARIA dialog focus contract: on open, focus
+  moves into the dialog — to a `[data-autofocus]` element if the consumer marked
+  one, else the first focusable control, else the panel itself; Tab and
+  Shift+Tab wrap within the dialog; on close, focus returns to the opener.
+-->
 <script lang="ts">
-  import type { Snippet } from 'svelte'
+  import { type Snippet, onMount } from 'svelte'
 
   import X from '@lucide/svelte/icons/x'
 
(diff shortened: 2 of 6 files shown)


agazso (approved):
Two comments:
- FOCUSABLE_SELECTOR doesn't filter invisible elements
- On mobile it's better to not autofocus to text input or textarea when the dialog appears, because it also brings up the keyboard and too much movement can be confusing/bad UX

agazso (approved):

Issue #392, opened 2026-07-02
New UI: resize partial failure leaves the drive diluted with no in-app resume


vojtechsimetka:
## Plain-language summary

A *drive* is paid storage on Swarm: it has a **size** (how much you can store) and a **lifespan** (how long the network keeps the data — it runs on a prepaid deposit that depletes over time).

The **Increase size** dialog offers a "Keep current lifespan" toggle (on by default). Growing a drive spreads its remaining deposit over more storage, which shortens the lifespan — doubling the size roughly halves the time left. To keep the lifespan, one press of **Proceed** therefore performs two operations in sequence:

1. **Grow the drive** — free, but the lifespan drops proportionally
2. **Top up the deposit** — paid, restores the lifespan

## The problem

If step 1 succeeds but step 2 fails (node/network hiccup), the dialog shows a generic error with **Try again** / **Close**:

- **Try again** works well: the dialog remembers it only owes the top-up, locks the form inputs, explains what's pending, and retries just step 2.
- **Close** silently drops the plan. The user chose "keep current lifespan" but ends up with a drive whose lifespan quietly shrank — e.g. 10 months left became ~5 — and nothing in the UI says so afterwards, remembers it, or offers to finish the job.

After closing:

- The drive list shows the new, bigger size and the shortened lifespan as if that were what the user asked for.
- Re-opening *Increase size* starts fresh from the new size and only offers even bigger sizes — there is no way to say "just finish the lifespan top-up".
- The only recovery is the separate **Extend lifespan** dialog, where the user must work out on their own how much time to add back. The app never tells them a top-up is missing or how large it was.

No money is lost — the failed step is the payment, so the user hasn't paid for the missing lifespan. But a payment flow that silently under-delivers on the choice the user made is a trust problem.

## What design should decide

1. **The error screen** — today it shows a generic error message plus Try again/Close. Should it spell out the half-done state ("the size increase went through; the lifespan top-up didn't") and the consequence of leaving ("this drive now expires ~X sooner")? Should there be a third path, e.g. "finish this later"?
2. **After the user leaves anyway** — should the app remember the unfinished top-up and surface it on the drive itself (e.g. a warning state on the drive row with a "complete top-up" action), or is steering the user into *Extend lifespan* — ideally pre-filled with the missing amount — enough?
3. **Copy** — this is a payment context; the wording needs to be precise about what happened, what it means, and what fixing it costs.

[...]

PR #511, opened 2026-07-31
feat: node-less drive extend and resize with on-chain payments

### LINKED ISSUE #309: Full postage stamp management in swarm id
- [x] buy
- [ ] top up
- [ ] dilute

These could be added through widget if available before end of MVP

vojtechsimetka:
Buying a drive, extending its lifespan and increasing its size all run in-app now — no Bee node owning the batch, and no fund.bzz.limo popup. The account's derived postage signer signs Gnosis transactions directly against the PostageStamp contract, and money reaches it through payment screens we own.

Implements [docs/Postage-On-Chain-Engine.md](https://github.com/snaha/swarm-id/blob/feat/onchain-drive-payments/docs/Postage-On-Chain-Engine.md) and [docs/Drive-Payment-Flow.md](https://github.com/snaha/swarm-id/blob/feat/onchain-drive-payments/docs/Drive-Payment-Flow.md), both in this PR.

## What shipped

- **Buy, extend and resize are node-less.** `runPurchase`/`runExtend`/`runResize` go through one funding seam, so all three inherit the same payment screens, rail and bundling. The derived signer buys what it owns — no throwaway creator wallet, no ownership handover, no dust.
- **Order is load-bearing.** `increaseDepth` is owner-only and checks the post-dilution per-chunk balance against its ~24h floor **before** any compensation, so the node-era dilute-then-top-up order reverts exactly in the common keep-lifespan case. This inverts it — top up first — and plans from the *live* remaining balance, never the stored snapshot.
- **One atomic transaction** where the EIP-7702 delegate exists, which includes Gnosis mainnet: extend is `approve + topUp`, resize is `approve + topUp + increaseDepth`. Chains without it fall back to sequential calls. Atomicity does not relax the ordering — a bundle that dilutes first reverts identically.
- **Payment is a swappable rail.** Relay Protocol in production. Relay cannot run locally at all, so a dev rail stands in: a genuine deposit on a second local chain, filled by a separate solver process.
- **Nothing funds itself, and no rail is an error rather than a free top-up.** Two dev shortcuts are gone. The local rail used to mint the connected wallet 10 ETH on its way past, so "the wallet cannot afford this" was a state the setup could never reach. The funding seam used to answer a `FundingNeed` with a silent faucet transfer whenever no rail resolved, settling operations a production build refuses — money arriving with no screen, no signature, no record of who paid. Both are now what they are in production: an error at the wallet, and an error at the seam. A local run funds the wallet or the signer **out of band** first, from the `/dev` faucet.
- **A faucet worth the name on `/dev`.** Any address on either chain, any of the three assets — ETH on the fake mainnet (minted; there is no faucet account to transfer from), xDAI and BZZ on Gnosis (transferred from the baked faucet). It shows what the recipient and the faucet hold before and after, so out-of-band funding is a visible act rather than a guess.
- **A delegated signer is not a plain EOA, and transfers now price for it.** Once a signer runs a bundled operation it carries a 7702 delegation, so a native transfer to it executes the delegate's fallback and costs more than 21 000 gas. `transferNative` hardcoded 21 000 and reverted — which meant **the local solver silently failed to deliver** to any batch owner that had already bought once. It estimates now.
- **Paying from Gnosis skips the bridge entirely.** The destination is the source there, so the wallet sends xDAI straight to the batch owner and the operation continues into the same swap — no bridge fee, no solver, no third party between signing and delivery. It used to go through Relay, which paid a cross-chain network to move money across zero chains. It is also the only route that is genuinely the same code locally.
- **One local chain, committed to the repo.** A Gnosis mainnet fork carrying the real BZZ market and SushiSwap pool, with the Swarm contracts deployed on top **at their mainnet addresses**. The production path runs against it unmodified, offline.
- **Genesis hash tells it from mainnet.** The local chain reports chain id 100 deliberately, so an app resolving contract addresses finds what it expects; chain id therefore cannot distinguish them and genesis can. Network settings, the `/dev` banner and rail selection all hang off that.
- **One cluster, from the `vendor/bee-compose` submodule.** `pnpm dev:bee*` and the published `@snaha/bee-compose` are gone — they expected the old DEX-less chain and crashlooped the queen against this one.

## Running it locally

Everything below is offline. You need Docker and a browser wallet — no funds, no testnet, no account anywhere.

The chain snapshot and the cluster live in a submodule, so clone with it:

```bash
git clone --recurse-submodules …     # or, in an existing clone: git submodule update --init
pnpm install
pnpm dev:local                        # cluster + both chains + solver + UI + demo
```

The first run compiles Bee from source (a few minutes); later runs take seconds. Wait for `local solver: watching for deposits to 0x…`.

| | |
| --- | --- |
| Identity UI | http://localhost:5500 |
| Demo | http://localhost:3500 |
| Queen Bee API | http://localhost:1633 |
| Gnosis Chain (fake), id 100 | http://localhost:9545 — **what the wallet pays from** |
| Ethereum Mainnet (fake), id 31337 | http://localhost:31337 — only for the bridged route |

### MetaMask

**The easiest way is `/dev` → Chain tab → Wallet networks**, which has a button per chain. Failing that, the payment flow offers the chain itself: connecting a wallet prompts MetaMask to add and switch to it.

**Add *Gnosis Chain (fake)* and pay from that.** It needs no bridge and no solver — the wallet simply sends xDAI to the batch owner. *Ethereum Mainnet (fake)* is the bridged route, and that one does need the solver running.

**Fund the wallet you connect with — nothing tops it up.** The dev rail used to mint the payer 10 ETH on its way past. It no longer does: no production rail can hand the payer money, and while one did, "the wallet cannot afford this" was the one state the whole setup could never reach. An empty wallet now fails on send, where it would live. Either:

- **`/dev` → Chain tab → Faucet.** Any address, any of the three assets, on either chain — ETH on the fake mainnet, xDAI and BZZ on Gnosis. Out-of-band funding, the way a testnet faucet works.
- **Import anvil's first key**, which starts out holding 10 000 native on each chain — 10 000 ETH on the fake mainnet, 10 000 xDAI on Gnosis — and no BZZ:

```
Private key  0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
Address      0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
```

That is the publicly known anvil/hardhat account #0. It is worthless outside a local chain — never put anything real behind it.

Adding the networks by hand instead:

| | Gnosis Chain (fake) | Ethereum Mainnet (fake) |
| --- | --- | --- |
| RPC URL | `http://localhost:9545` | `http://localhost:31337` |
| Chain ID | `100` | `31337` |
| Currency | xDAI | ETH |
| Needs the solver | no | yes |

> **`RPC endpoint … consider using a different RPC endpoint` (code -32002).** MetaMask rejects an endpoint whose latest block looks too old, and a local chain only mines when something writes to it — so after an idle spell or a fresh bake it is refused before the app is reached. Any write unblocks it; the `/dev` faucet is quickest.

> **After `pnpm dev:local:fresh`**, the chains restart at block 0 while MetaMask still holds the old nonces, and transactions fail with *"nonce too high"*. Fix: MetaMask → Settings → Advanced → **Clear activity tab data**.

### The walk-through

1. **http://localhost:5500** → **Get started**, create an identity (password is quickest), then **Stay local for now**.
2. **Settings → Network settings → Use local → Save.** It probes and picks whichever local chain is answering.
3. **Storage → Add drive** → pick a size → **Proceed**. This is a real purchase: a new account's postage signer holds nothing, so it raises a funding need and the payment screens open. (Fund that signer from the `/dev` faucet instead and no screen opens at all — nothing is owed. That is how the drive e2e suites buy.)
4. **Connect wallet → MetaMask**, and accept the prompt to add and switch to the chain. Connecting offers it, rather than waiting until you press Pay, so the balance you funded above is visible while you review the quote.
5. Leave the chain on **Gnosis Chain (fake)** — the route with no bridge. Review the quote → **Pay with your wallet** → approve.
6. The drive appears once the transfer confirms.

What just happened: your wallet sent xDAI to the drive's batch-owner address, that xDAI was swapped for BZZ through a real SushiSwap pool, and the batch was created on the real PostageStamp contract — as a single atomic EIP-7702 transaction.

Switch the chain to **Ethereum Mainnet (fake)** to take the bridged route instead: the wallet signs a deposit on one chain, the solver process sees it and pays out on the other, and the `[solver]` lines in your terminal report the delivery before the rest proceeds identically.

For extend and resize: expand the drive → **Extend lifespan** → switch to **years** and bump it to 3. A small extend is covered by the batch's leftover funds and never asks for payment, which is the usual reason the payment screens seem not to appear.

The chain is stateful and nothing rewinds it — every purchase trades against a real, thin BZZ pool. `pnpm dev:local:fresh` restores the baked snapshot (and wipes node state, so drives created earlier will point at batches that no longer exist).

## Screens

Captured from the running app (`tests/screens.capture.ts`). Extend and resize are **real transactions**; the payment screens stub Relay's quote and inject a wallet, since its API is unreachable from a test run and a headless browser has no wallet. The components and their data flow are the production ones.

| Extend | With cost | Increase size | Keep lifespan |
|---|---|---|---|
|  |  |  |  |

| Payment method | Pay | Working | Done |
|---|---|---|---|
|  |  |  |  |

Checked frame by frame against the [STORAGE section](https://www.figma.com/design/1UCqrAPBp5jBG4IXjp4bEg/SwarmID-MVP--Copy-?node-id=127-9470). Two deliberate deviations, both on the pay screen: no per-token "N available" balance (it needs a per-chain balance read for every listed token, adding RPC failure modes to the quoting path), and the cost breakdown is always visible rather than behind a chevron (there are one or two rows).

## Tests

Everything below runs in CI.

| Suite | | Covers |
| --- | --- | --- |
| unit | 53 lib + 4 multichain + 13 ui files | planners, funding maths, ABI selectors, rail quoting |
| `test:fork` | 11 | real pool and contracts: swap → create → topUp → increaseDepth, 7702 bundles, non-owner revert |
| `test:e2e` | 32 | the app, incl. buy/extend/resize on chain and a full payment through the rail |
| `test:cluster` | 2 | the nodes ingest a batch bought through the multichain path, and a client-stamped chunk round-trips |
| `test:live` | 6 | Relay's real quote API — our request shape and response mapping |

`payment-rail.test.ts` is the one that exercises the payment screens end to end: a real deposit, a real solver fill, the real Sushi swap and a real `createBatch`, asserting the batch on chain is owned by the account's own signer. CI starts the source chain and the solver for it.

Every other drive suite used to pin the rail OFF, which selected the silent faucet settlement described above — so eight purchases across four suites were covered by a code path production does not have. With that path gone, they fund the postage signer **out of band** before buying (`fundPostageSigner`, the same chain faucet a developer uses by hand, sized from a measured purchase: a depth-17 one-year drive costs ~2 BZZ). The operation then finds the funds already at the owner address and no payment screen opens — because nothing is owed, not because a rail was suppressed. The suite that asserts a *failed* purchase deliberately does not prefund.

## Known gaps

- **Relay's delivery is unexercised.** Its quote is contract-tested against the live API, but no local chain can take a solver payout. One production canary before release — a small real extend against a throwaway batch.
- **On chains without the 7702 delegate**, a resize whose depth increase fails leaves the lifespan longer and the size pending; retrying re-prices against the now-longer lifespan and charges again. Unreachable on mainnet, where the delegate exists. Documented in Drive-Payment-Flow §6.6; resuming it properly needs the target depth persisted on the record.
- **The `/dev` route still ships** its own lazily-loaded chunk carrying dev-chain helpers, including a public faucet key. It is out of the entry bundle, but whether that page belongs in a deployment is a separate question.
- The **"Lifespan decreased"** Figma modal (`481:14324`) describes the old ordering and is unreachable; copy needs flipping (#392).

Closes #309
Touches #392


vojtechsimetka:
Closing as this is being refactored into more manageable PRs

PR #534, opened 2026-08-14
feat(dev): rehearse the cross-chain payment locally, behind the rail seam

vojtechsimetka:
**Stack 5/6** — base `stack/3-onchain-payments` ([#533](https://github.com/snaha/swarm-id/pull/533)).

Relay cannot run locally at all: it is an intent/solver network, so quotes come from a hosted API and deliveries from off-chain solvers paying out on real Gnosis. This adds the stand-in — a bare anvil on 31337 the wallet signs a real deposit against, and `multichain/src/local-solver.ts` watching it and paying out from the Gnosis-side faucet. The deposit carries its delivery instruction in its own calldata, so the solver is stateless; stop it and a payment hangs and then fails, which is what a solver outage looks like.

That is what makes the `method → connect → configure → wallet-approval` half of the flow exercisable at all.

**It rehearses the experience only.** The prices are invented, the step list is shorter than Relay's (an ERC-20 source needs an approve first), and the failure and refund behaviour is nothing like it. A green local run must never be read as *"the payment path works"*. Paying from Gnosis needs none of this and is the same code locally as in production — it stays the quickest way to try a payment.

## The rig must not ship

`$lib/payment/dev-funding` is the only door production code uses, and `vite build` swaps it for an inert stub — a `pre` plugin, because SvelteKit's `$lib` alias resolves first. A dead `import.meta.env.DEV` branch does **not** remove a static import, and these modules have top-level side effects, so Rollup keeps them regardless.

Reading the source cannot verify this, and it has been silently broken twice. So CI greps the built assets after `pnpm build`: `anvil_setBalance` may appear only in the `/dev` route's own chunk.

`pnpm dev:local` brings up the cluster, both chains, the solver, the UI and the demo together.

## What is left here after the split

The `/dev` rebuild and the faucet moved to [#540](https://github.com/snaha/swarm-id/pull/540); this PR adds back only the parts that need a source chain — the faucet's ETH leg, the source-chain field in Network settings with its own `Connected to: …` line, and the `Ethereum Mainnet RPC (fake)` status row. The **Use local** button went with the faucet, since pointing at the local chain is useful a PR earlier.

The **Gnosis** endpoint's `Connected to: …` line went with it too, for the same reason: it answers "which chain is this" about the endpoint #540 already teaches the app to identify, so it belongs beside that PR's banner and **Use local** button rather than two PRs later. What is left here is the second field, which only exists once there is a source chain to point it at — and the `Gnosis Chain RPC` label, which only reads correctly once a second RPC sits under it.

## And the suite that proves it

`ui/tests/payment-rail.test.ts` lands here, with the rail it exercises: method → connect → configure → pay, then a real deposit on the source chain, the solver's delivery, the real Sushi swap and a real `createBatch`. It is the one suite that deliberately does **not** prefund, so the payment screens actually open. The Playwright job gains the source chain and the solver for it.

## Verification

`pnpm check:all` green. E2E run locally as this PR's CI does, with the source chain and solver up: **29 passed, 0 failed** (2 skipped — the cluster pair), including both `payment-rail` tests. The bundle guard verified by hand on a real `pnpm build`: no dev code outside the route chunks.


agazso (commented):
Reviewed against the branch tip. This is careful work — the dev seam's design, the two bounded legs, and a live contract suite that derives its pairs from the picker's own table are all the right shapes. Six findings; the first two are provable from the branch alone.

## Blocking

**1. Two different tokens, both labelled "USDC (USD Coin)", in the Gnosis picker**

- `relay.ts` — `PAYMENT_TOKENS[gnosis.id]` lists `0x2a22f9c3b484c3629090feed35f17ff8f88f76f0`
- `gnosis-direct.ts:109` — builds its USDC from `gnosisMainnetSettings().addresses.usdc`, which is `0xDDAfbb505ad214D7b80b1f830fcCc89B60fb7A83` (`multichain/src/settings.ts:66`, and this branch's own `gnosis-direct.test.ts:256`)

`combineRails` dedups by **address**, so both survive into `tokens(100)` and the selector shows two identical-looking rows: one served by the direct rail, one by Relay, on different token contracts. `0x2a22…` is USDC.e; `0xDDAf…` is the token the rest of this codebase transacts in and the one the swap side knows how to accept.

A holder of `0xDDAf…` who picks the Relay row sees no balance and concludes they cannot pay; a holder of USDC.e who picks the direct row gets a rail that cannot serve them. Whichever is intended, they cannot both be "USDC (USD Coin)".

**2. `resolve-rail.ts`'s composition doc contradicts the code it documents**

The module comment says *"the direct rail moves native xDAI because it is a plain transfer"* and *"the direct rail takes the native token, Relay keeps the rest"*. But `ACCEPTED` (`gnosis-direct.ts:96`) is **xDAI, WXDAI, USDC, BZZ**, and `tokens()` returns all four (`:433`). The direct rail claims four tokens on Gnosis, not one.

The bug the description names — a chain-level claim silently removing Gnosis USDC — is real, and per-token dispatch is the right fix. But the rule as stated is not the rule that ships, and finding 1 is what that gap produces.

## Should fix

**3. The bundle guard can pass while dead** — `.github/workflows/ci.yml`

The check asserts (a) exactly one `/dev` chunk carries `swarm-id-dev-source-rpc`, then (b) that no *other* file matches `anvil_setBalance|localhost:31337|a0ee…9720`. Nothing asserts the canaries still **exist**. Rename a constant, change the port, rotate the faucet key, and (b) passes vacuously — green forever, guarding nothing.

The failure mode has already arrived once: the step's own comment explains that `anvil_setBalance` tree-shakes away, i.e. one of the three canaries is inert today. That is the check telling us it can be inert without saying so.

Assert each canary matches at least once inside `$dev_chunk` before searching for it elsewhere. Then a canary that stops existing fails loudly instead of silently — the same rule as the lesson already in `docs/Account-Bus-Priorities`: a negative assertion is only as good as proof the channel is live.

**4. `EXECUTE_TIMEOUT_MS` counts human signing time** — `relay.ts`

`withTimeout` wraps the whole `execute`, and the comment says so ("counted from the wallet prompt"). Ten minutes is generous for a delivery but not for a person on a mobile wallet — app switch, biometric, a network prompt, a second device. When it fires, a payment that is progressing normally is reported as failed, with money in flight.

`onProgress` is already wired. Resetting the deadline on each step makes it time out on an SDK that has stopped reporting — the stated purpose — rather than on a slow human.

**5. Production offers the bridged rail whatever chain the app is pointed at** — `resolve-rail.ts`, `relay.ts`

`resolveBridgedRail` returns `relayRail` unconditionally outside DEV, and `quotePayment` hardcodes `toChainId: GNOSIS_CHAIN_ID`. The Gnosis endpoint is user-editable in network settings, so a shipped build pointed at a fork or testnet gives: direct rail gone (identity `unsupported`), Relay offered, xDAI delivered to **real** Gnosis, and a shortfall the app can never see clear because it is reading a different chain.

The doc bullet ("a production build always pays for real, whatever chain it is pointed at") answers a different question — that one is about not faking a payment, not about delivering to a chain the app is not looking at. Requiring a proven-mainnet identity for the bridged rail in production too would close it.

**6. `delivers.amount` is the requested figure, not the promised one** — `relay.ts`

`quotePayment` returns `request.xdaiWei - request.gasXdaiWei`, computed from the ask and never from the quote's `minimumAmount`. `swapDelivered` later swaps exactly that figure.

The live suite pinning `minimumAmount === requested` is the right guard for today. But if Relay's fee model ever makes a delivery land short, the swap runs against a balance that cannot cover it — **after** the user paid — and `ensureFunded`'s shortfall re-check happens later, so it surfaces as a broken operation rather than the re-price the retry path was built for. Flooring the swap at the owner's actual balance (less the gas share) turns a future drift into a retry instead of a failure with money already moved.

## Minor

- `readBalances`: the wallet path has no timeout while the RPC path has `BALANCE_TIMEOUT_MS`. Fire-and-forget, so it only leaves a promise pending, but the asymmetry looks unintended.
- The provider-events `$effect` cleans up through `removeListener?.` — a provider with `on` but no `removeListener` leaks handlers across provider changes.
- `onStatus?.(currentStep.action)` puts Relay's own wording on our screen verbatim, days after #617 landed a PR about that class of text.

## Process

Seven commits behind `main`, and `payment-dialog.svelte` is touched by both this and #629 — the same merge-order hazard that bit #601/#610, in the money screen. The overlap is small (the `$state.raw` quote holder vs an `initialMethod` prop) but it is in the same hunks, so whoever lands second rebases deliberately rather than trusting a clean auto-merge.

vojtechsimetka:
Thanks — all six addressed, plus the three minors and the rebase. One I could not reproduce as described; details at the end.

## Blocking

**1. Two "USDC (USD Coin)" rows on Gnosis** — confirmed exactly as described: `combineRails` dedups by address, so `0x2a22…` (Relay) and `0xDDAf…` (direct) both survive into `tokens(100)`.

Kept both, named apart. They are genuinely two contracts with two holder bases, and each rail can carry the one it lists — Relay converts USDC.e to xDAI, the direct rail transfers the Omnibridge token the swap side already knows. Dropping either would strand its holders. So Relay's row is now `USDC.e (Bridged USD Coin)`, as a wallet names that contract, and a test in `relay.test.ts` pins the two apart by both address and symbol.

**2. `resolve-rail.ts`'s composition doc** — confirmed. Rewritten to the rule that ships: the direct rail wins the Gnosis assets with a route to BZZ (xDAI, WXDAI, USDC, BZZ), Relay keeps the rest. Same correction in `docs/Drive-Payment-Flow.md` and in the `combineRails` test's own comment, which had inherited the wrong rule.

## Should fix

**3. The bundle guard could pass while dead** — confirmed, and it is worse than the finding says. Grepping a real `pnpm build`:

```
anvil_setBalance                          → build/_app/immutable/nodes/12.D-LvGdRO.js
localhost:31337                           → build/_app/immutable/nodes/12.D-LvGdRO.js
a0ee7a142d267c1f36714e4a8f75612f20a79720  → (nothing, anywhere)
```

The inert canary is the **solver address** — the one the step's comment singles out as unable to tree-shake away because "the rail is built out of it". `/dev` imports the faucet half of `local-payment-rail` and nothing else, so the rail's own execute path is shaken out of the dev chunk too. It has been matching nothing.

That also rules out the fix as proposed: asserting each canary appears inside `$dev_chunk` fails today, correctly-configured. Absent from the bundle is the *right* state for that one. So liveness is asserted against the **declaration** instead — anchored on `LOCAL_SOLVER_ADDRESS = "0xa0ee…"`, not on any mention, so a comment naming a renamed helper cannot keep a canary looking alive. Verified both ways: green as-is, red when I rotated the key.

**4. `EXECUTE_TIMEOUT_MS` counting human signing time** — confirmed and fixed. Added `withIdleTimeout` beside `withTimeout` in `lib/src/utils/promise.ts` (a hand-rolled resettable race is exactly what `AGENTS.md` forbids), which restarts its clock on each `keepAlive()`. `relay.ts` pings it from `onProgress`, so ten minutes now bounds *silence*. Tests: a payment reporting every nine minutes survives an hour; one that stops reporting dies a full deadline after its last report.

**5. Production offered Relay whatever chain the app is pointed at** — confirmed and fixed. `resolveBridgedRail` now requires a proven-mainnet identity in every build, not only in a dev one. A shipped build pointed at a fork gets the direct rail alone.

## Minor

- `readBalances`' wallet path now carries the same `BALANCE_TIMEOUT_MS` the RPC path had.
- The provider-events effect subscribes only to a provider that also has `removeListener`, so nothing can be registered that cannot be torn down.
- `currentStep.action` is gone. The card is driven by the step `id` and the item's `progressState` — both typed — mapped to this app's own wording, matching what the local rail already says. An unknown step is described, not quoted; a `signing` state says the wallet is holding the request.

## Not reproduced — 6, `delivers.amount`

The finding's chain is `quotePayment` → `delivers.amount` → `swapDelivered`. That middle link does not exist: `settleWith` (`funding.ts`) returns the quote **untouched** when `delivery.input === 'xdai'`, and Relay always delivers xDAI. `payment-dialog.svelte` reads only `delivers.input` on that path. So `swapDelivered` spends `quoteFunding`'s own `xdaiForBzzWei`, never the rail's figure — `delivers.amount` is inert for both bridged rails, and the `settleWith` doc explains why deliberately (the xDAI swap input also holds the owner's pre-existing surplus, which no rail carried and none can see).

The residual risk is real but much narrower than stated: an under-delivery of δ eats the gas budget rather than the swap, so small drift surfaces as the re-price the retry path was built for, and only δ larger than `GAS_BUDGET_XDAI_WEI` reaches a reverting swap. Flooring at the live owner balance would mean re-reading it between payment and swap on the most delicate money path, against a drift the live suite's `minimumAmount === requested` pin currently rules out on all twelve pairs. Left alone — happy to do it if you'd rather have the belt.

## Process

Rebased onto `main` (was ten behind). The `payment-dialog.svelte` overlap with #629 auto-merged, but the hazard was real and one hunk further out: #629 adds an **up-front** method chooser, so `addDrive` now answers it and `PaymentDialog` opens on `initialMethod="built-in"`. The e2e's step 1 was re-picking a method already chosen — which would have hidden an `initialMethod` that stopped reaching this path. It now asserts the choice carried instead.

`pnpm check:all` green at the tip, on a freshly built `lib/dist`. The Playwright payment-rail suite has **not** been re-run against the local stack after the rebase — that e2e change is reasoned from the merged source, not measured.

agazso (approved):
Reviewed the full diff (41 files, ~3.9k lines) — production rail, dev rig, guards, tests. The money path holds up. Below is what I'd still push back on, ranked.

## Worth fixing

**1. On a Relay ERC-20 route the approve prompt is announced as "the payment".** `relay.ts:231` short-circuits on `progressState === 'signing'` *before* consulting `stepId`, so `STEP_STATUS.approve` ("Confirming the approval") is only ever reachable in `confirming` — never in the window where the wallet is actually holding the approve. Five of the twelve picker pairs (mainnet/Base/Arbitrum/Optimism/Polygon USDC) sign approve-then-deposit, and both prompts read "Confirm the payment in your wallet".

Compounding it: `payment-dialog.svelte:203` computes `separateGasLeg` from `delivers.input !== 'xdai'`, which is *never* true for a Relay quote — so the "this payment is two transactions" hint at line 638 is structurally unreachable on the one rail where Relay itself needs two. The direct rail bothers to warn; the local rehearsal words it correctly ("Confirming the approval"); production is the only place that gets it wrong, which is why nothing caught it.

`relay.test.ts:132` walks `approve`+`confirming` and `some-future-step`+`signing` but never `approve`+`signing` — the exact combination. Fix is one line in `stepStatus` plus widening the two-prompt hint to any quote whose steps include an approve.

**2. The live suite's retry budget can exceed its own timeouts, in precisely the scenario it was added for.** `beforeAll` is sequential; a refused pair costs up to `2 × (15s delay + 25s request)` on top of the first attempt. Six refused pairs — Relay degraded but answering, so `refused`, not `unreachable` — is >300s, past `hookTimeout` (`vitest.live.config.ts:28`), and the job is capped at `timeout-minutes: 10` including install. The outcome is an opaque hook timeout that names no pair, rather than the "this route is gone" report the design wants. Cap the total retry budget, or do retries in a second pass over only the refused pairs.

**3. An unreachable pair passes green.** `relay.live.test.ts:227` returns early on `status === 'unreachable'`, and `unreachable` includes the 25s `AbortSignal.timeout` — so a route that *hangs* is indistinguishable from being offline, and its `it` passes having asserted nothing. Only Base ETH is covered by the CI reachability check at line 158. `ctx.skip()` would keep those honest without failing an offline developer.

**4. `network-settings-dialog.svelte:105` persists the source RPC unvalidated.** `canSave` gates only `beeNodeUrl`/`gnosisRpcUrl`; `source?.saveRpcUrl(sourceRpcUrl.trim())` writes whatever was typed. A scheme-less `localhost:31337` saves fine and then silently removes the bridged rail (`probeChainId` throws → `resolveLocalRail` → `undefined`) — the same failure the Gnosis field's own comment says its validator exists to prevent. Dev-only, but it's the `isValidUrl` already sitting above it.

## Doc/comment accuracy

**5. `EXECUTE_STALL_TIMEOUT_MS`'s rationale doesn't survive the signing window** (`relay.ts:36-54`). It argues a total deadline is wrong because it "would count the time a person spends in their wallet — an app switch, a biometric prompt, a second device". But the SDK reports `signing` once and then goes quiet until the user returns, so wallet time *is* counted — just re-based per step. 10 minutes is generous enough that this is a wording problem, not a bug; the comment should say "per step" rather than claim the case away.

**6. Dangling comment at `gnosis-direct.ts:45-49`.** `GENESIS_PROBE_TIMEOUT_MS` moved to `payment-rail.ts:288`; its doc block stayed behind and now floats above `GNOSIS_NAME`, describing nothing — while the constant at its new home has no doc at all.

**7. `local-payment-rail.ts:430`'s `get chains()`** claims per-access reads make a settings edit take effect without a reload. `combineRails` (`resolve-rail.ts:40`) snapshots `chains` into a plain array, so the getter buys nothing; the behaviour survives only because `resolvePaymentRail()` re-runs per payment. Either the comment or the getter should go.

## Noise, take or leave

- **Three wallet genesis probes per direct payment**: `switchWalletChain` probes in `followChain`, again in `pay()` (`payment-dialog.svelte:548`), then `executeDirectPayment` does its own (`gnosis-direct.ts:342`). The last is a deliberate backstop; the middle two are back-to-back and one could be handed down.
- **Two `readBalances` per chain change** (`payment-dialog.svelte:712-713`): `followChain` and `refreshQuote` each fire one, reading from different sources (wallet vs endpoint), whichever lands last winning. Same numbers, so harmless — but undocumented, and it doubles the RPC fan-out on every selection.
- **`walletClientFor`'s `defineChain` fallback** (`relay.ts:189`) is unreachable: chains only ever come from `WALLET_CHAINS`. Dead defensive branch that invents `nativeCurrency: Ether` for an arbitrary id.
- **The local solver never checks a native deposit's value** against `instruction.xdaiWei`. The token path carefully rounds up so a pull can't be zero (`local-payment-rail.ts:169`), while the native path truncates down and nothing rejects a zero-value deposit. Not reachable at realistic amounts, but the asymmetry contradicts the "never fake a payment" rule the pull path is written to.

## What I checked and found sound

The bundle guard is genuinely airtight for its threat model: a production route importing `$lib/dev/*` either hoists the marker into a shared chunk (`found` ≠ 1) or lands it in that route's own node chunk (`found` = 2) — both fail. The source-side canary staleness check closes the "negative assertion about a string nobody spells any more" hole properly. `withIdleTimeout` handles the sync-throw, post-settle-`keepAlive` and timer-clearing paths correctly, and the tests cover all three. `combineRails`' per-token dispatch fixes a real stranding bug, and the `USDC`/`USDC.e` split is correct against `settings.ts:66`. `switchWalletChain`'s genesis verify → re-offer → worded refusal chain is well tested, including the MetaMask "network already exists" path.

Approving — none of the above blocks the merge; (1) is the one I'd take before it ships.

vojtechsimetka:
Thanks — took four, corrected three comments, filed three, and pushed back on one. `fad8cccad`.

## Taken

**1 — the reachable half.** The stated mechanism doesn't hold: `approve` is a **transaction** step, and the SDK sets `progressState = 'confirming'` *before* it prompts (`transactionStep.ts:15`, ahead of `sendTransactionSafely`). Only `handleSignatureStepItem` ever sets `signing`. So `approve` + `signing` is unreachable, and `confirming` **is** the window where the wallet holds the approve — "Confirming the approval" already covers it, which is what `relay.test.ts:132` pins.

But there is a real one right next to it, and it is worse than the wording you found: on a wallet that batches atomically (EIP-5792), `canBatchTransactions` collapses `[approve, deposit]` into one step and `prepareBatchTransaction` renames it **`approve-and-deposit`**. That id matched nothing, so the card fell through to "Cross-swap xDAI on Relay" — the delivery leg — while the wallet was still holding the request. Fixed by wording a batched step as the step it ends in; the single signature *is* the payment. Test added.

On `separateGasLeg`: right that it's structurally unreachable for Relay, but the hint reads "approve **the gas** in your wallet first, then the payment itself" — that's the direct rail's gas leg, and it isn't what Relay's approve is. Widening the condition would put wrong words on the right screen, so I left it; the batched-step fix covers the case where Relay's two prompts actually become one.

**3 — unreachable pairs.** Fixed with `context.skip()`. You're right that a hang and an outage are indistinguishable there and a green tick reads as coverage the suite doesn't have.

**4 — unvalidated source RPC.** Fixed with the `isHttpUrl` already imported two lines above, plus the inline error the other two fields have.

**5, 6, 7 — comments.** All three overstated the code:
- `EXECUTE_STALL_TIMEOUT_MS`: you're right, wallet time is still counted. Reworded to say the clock is re-based *per step*, so two prompts get a budget each rather than sharing one — which is the actual claim.
- The `GENESIS_PROBE_TIMEOUT_MS` block moved to the constant, which had no doc.
- `get chains()`: reworded. The getter stays, though — it isn't only about reload behaviour, it's what keeps a `localStorage` read out of module load. What was wrong is the "without a reload" half, now stated as "on the next `resolvePaymentRail()`, once per payment, not live".

## Filed

- **#637** — the live suite's retry budget. Confirmed the arithmetic: 105s per refused pair worst case, so **three** refused pairs already exceed `hookTimeout: 300_000`, not six. Worth noting the asymmetry too — `unreachable` short-circuits without retrying, so only the degraded case is slow, which is the one the retries are for. Second-pass-over-refused-pairs looks like the right shape.
- **#638** — the duplicate reads. Verified both: three genesis probes per direct payment (`followChain` → `pay` → `executeDirectPayment`, the last deliberate) and two `readBalances` per chain change from different sources, last-write-wins. Folded the dead `defineChain` fallback in there.
- **#639** — the solver's unchecked native deposit value. Agreed the asymmetry against the carefully-rounded pull path is the point.

## Verification

`pnpm check:all` green. `pnpm test:live` re-run against the real Relay API after the skip change: **all twelve picker pairs still quote and route** (17 passed, 1 CI-only skip, 12s).

Still standing from the last round: the Playwright payment-rail suite has not been run locally since the rebase — CI's run on the branch is the first real exercise of that change.

PR #537, opened 2026-08-14
feat(lib)!: multi-batch accounts — batch listing, upload targeting, account-scoped write lease

agazso:
> [!WARNING]
> **Breaking changes.** `getPostageBatch()` is **removed** — use `getPostageBatches()` and pick the stamp you need. Internally, `BatchWriteCoordinator`'s deps/`withWrite` now take a per-write stamper, and the cross-tab write-lock key changed from `swarm-write-<batchId>` to `swarm-write-account-<accountId>`.

## Problem (design doc §2 — "Only One Stamp Per Account is Usable")

An account can hold several postage stamps (each shown as a "drive" in the identity UI), but the SDK only ever exposed and used the **single resolved** stamp — per-app override → account default. No API reached a non-default stamp, and uploads couldn't target one.

## 1. API: list batches + target uploads

- **Read** — `getPostageBatch()` → **`getPostageBatches(): Promise<PostageBatch[]>`**: every non-tombstoned stamp the account owns, live-enriched (`usable`/`exists`/`batchTTL`, bounded by a per-stamp timeout with snapshot fallback), with **`isDefault`** flagging the batch untargeted uploads consume.
- **Write** — **`UploadOptions.batchID`**: target a specific owned stamp on data/file/chunk uploads, SOC & GSOC writes, sequential-feed writes, `createFeedManifest`, and ACT uploads. Rejects a batch the account doesn't own. Epoch-feed writes still use the resolved stamp (their wire format carries no upload options — follow-up).

## 2. Account-scoped write lease (no more per-swap thrash)

The on-network partition claim (lock/intent/occupancy SOCs) is **account-scoped** — only the partition-*state* layer is per batch. The original targeting implementation rebound the whole stamper/coordinator per batch swap, paying a network lease release + full intent-guarded re-acquire (seconds) on every alternation. Now:

- `PartitionLease` keeps **per-batch** publish state and gains `joinBatch(stamper)`: a targeted batch seeds its counter for the held partition with **zero lock activity**.
- `BatchWriteCoordinator` is **one per account**: `withWrite(stamper, op)` takes the target batch's stamper; the lease is stamped under the resolved default batch; lease-loss paths fence every joined batch's stamper; teardown/idle-yield flush each joined batch's final counter before the release sentinel; the refresh tick heartbeats every joined batch's resume pointer.
- The cross-tab Web Lock is **account-keyed**, which also makes the proxy and the UI's background sync genuinely mutually exclusive (they previously used different per-batch keys while mutating the same account-scoped lease).
- The proxy caches stampers per batch — targeted writes never mutate the default binding. Fixes two cross-batch bugs: the `StampWorkerPool` was reused across batches on a size-only check (stamping under the wrong batch), and the stamper flush/broadcast was batch-blind.
- **No wire-format changes** — cross-device behavior with old-version peers is unchanged.

## 3. Demo: upload batch selector

Sidebar "Upload batch" dropdown (shown for 2+ batches): **Default** (pre-selected, proxy resolves) plus each owned batch. Passed as `batchID` by every demo upload operation; the selection falls back to Default if the batch disappears.

## Migration

```diff
-const batch = await client.getPostageBatch()
+const batches = await client.getPostageBatches()
+const batch = batches.find((b) => b.isDefault) ?? batches[0]
```

Upload to a specific stamp:

```typescript
await client.uploadData(data, { batchID: someOwnedStamp.batchID })
```

## Notes

- **Privacy:** `getPostageBatches()` exposes every drive to any connected app; access is origin-only (no per-app permission model). Accepted trade-off.
- **Epoch feeds** don't support `batchID` yet (wire format); sequential feeds, SOC/GSOC, and `createFeedManifest` do.

## Tests

`pnpm check:all` green across the monorepo. New coverage: client/proxy batch-listing round-trips (tombstones, `isDefault`, enrichment timeout fallback), upload targeting (cache hit, reject-unowned, failed-build isolation, promotion when no default resolves), `PartitionLease.joinBatch` (seeding, readFailed, per-batch publish-cache independence, per-batch heartbeats, cross-batch hydrate), and coordinator multi-batch behavior (join-once/re-join-after-demote, fan-out invalidate ordering, teardown flush-before-sentinel, account lock key, per-batch pools/flush).


vojtechsimetka:
## Review

The account-scoped lease design reads well and the `joinBatch` factoring is clean. Four findings, two of them in the slot-safety path.

---

### 1. `teardown()` never publishes a joined batch's final counter (high)

`lib/src/sync/batch-write-coordinator.ts:440`

The lease stamper's counter is captured up front (line 414, before the unbind), but each secondary's is read *inside* the deferred `void this.lock(...)` callback:

```ts
const secondaries = [...this.joinedSecondaries.values()]
void this.lock(async () => {
  for (const s of secondaries) {
    const counter = s.getLocalCounter()   // <- runs after the unbind below
    if (counter === undefined) continue
```

Lines 466-468 run synchronously right after scheduling it:

```ts
this.forEachBoundStamper((s) => s.invalidateLease())
this.forEachBoundStamper((s) => s.unbindPartition())
```

`unbindPartition()` sets `this.partition = undefined` (`batch-utilization.ts:1516`) and `getLocalCounter()` returns `undefined` whenever `partition === undefined` (`batch-utilization.ts:1324-1328`). `navigator.locks.request` always invokes its callback asynchronously, so in a browser every `counter` is `undefined` and the loop `continue`s — while `lease.release(localCounter)` still writes the release sentinel.

Effect: sign out after a targeted upload to batch B and B's final counter is never published. A peer's takeover resumes B at its last acked counter, re-issuing slots this device already used and evicting B's chunks. `releaseIdlePartition` (line 1178) gets this right — it publishes *before* the unbind.

The new test at `batch-write-coordinator.test.ts:1599` can't catch it: the stamper double's `getLocalCounter: () => new Uint32Array(8)` ignores `unbindPartition`, and the `withAccountWriteLock` mock (test line 30) invokes `op()` synchronously.

Fix: capture the counters alongside `secondaries`, before the unbind.

---

### 2. `ensureBatchJoined`'s `bindPartition` clears the lease-lost breaker (high)

`lib/src/sync/batch-write-coordinator.ts:373`

`refreshTick` calls `signalLeaseLost()` **off** the write lock (line 1052) and only then queues `await this.lock(() => this.finalizeDemote())`. Interleaving:

1. An upload with `batchID = B` (B not yet joined) enters `withWrite`, takes the account Web Lock, passes `ensureLease` / `ensureHeldForUpload`. `ensureLeaseStillValid` is throttled to once per `LEASE_REFRESH_MS`, so it usually performs no lock-SOC read.
2. The refresh tick confirms displacement and calls `signalLeaseLost()`. That fans `invalidateLease()` over `leaseStamper` + `joinedSecondaries` — B's stamper isn't in the map yet, so it is **not** armed. `finalizeDemote` blocks on our lock.
3. `ensureBatchJoined(stamperB)` finishes `lease.joinBatch` and calls `bindPartition`, which sets `leaseStale = false` (`batch-utilization.ts:1418`), then `setLeaseValidUntil(lease.leasedUntil)` — still in the future.
4. The upload stamps under the lost partition with no breaker armed — exactly the peer-slot corruption `signalLeaseLost` exists to prevent.

This hole doesn't exist on `main`: `bindPartition` only ran inside `acquire()`, which never runs while a lease is held.

Suggested fix: record a per-session lease-lost marker and re-check it (together with `this.partitionLease === lease`) after `joinBatch` resolves, aborting the write rather than binding.

---

### 3. `joinBatch` can zero-seed a secondary batch on the adopt fast path (medium)

`lib/src/sync/partition-lease.ts:1115`

`joinBatch` sources `holderLeasedUntilMs` only from `lastSeenLeasedUntil`, which is populated exclusively by `refreshFromSwarm()` (and cleared at its start, line 344). `BatchWriteCoordinator.acquire`'s adopt fast path (lines 608-648) builds a fresh `PartitionLease`, calls `hydrate` + `adoptIfLive`, and returns with **no Swarm scan at all** — so `lastSeenLeasedUntil` and `unreadableLocks` are both empty.

With `holderLeasedUntilMs` undefined, `readStatePointer` scans only `cur` and `cur - 1` (`partition-state.ts:319-332`); the full lease span is scanned only when it is set. And with `lockUnreadable: false`, a "no pointer" verdict is treated as conclusive (`partition-state.ts:531-541` doesn't fire) and returns a zero counter.

Sequence: D1 holds partition P and writes batch B → D2 takes over P in session S1 but never targets B → D2's tab reloads inside the lease TTL, so S2 re-adopts P from cache → D2 uploads with `batchID = B`. D2 has no persisted `syncedReference` for (B, P), D1's last pointer write for B is older than ~2 pointer epochs, so `joinBatch` returns zeros, `ensureBatchJoined` binds B's stamper at zero, and D2 re-issues every slot D1 acked on B.

Pre-PR this couldn't happen — targeting B rebuilt the coordinator and ran a full `acquire()` (hence `refreshFromSwarm`) for B.

---

### 4. Stale lock-key in the published docs (low)

`docs-site/src/content/docs/multi-device-postage-batches.mdx:443` still reads ``(`swarm-write-<batchId>`) so a proxy iframe and the UI never collide locally``. `docs/Postage-Batch-Partitioning.md` and `docs/BatchWriteCoordinator.md` were updated to `withAccountWriteLock` / `swarm-write-account-<accountId>`; this one wasn't. The stale wording also hides the behaviour change that writers to *different* batches of one account now exclude each other.

---

### Checked and fine

`getPostageBatches` enrichment / per-stamp timeout / price memoisation; the `fetchBatchTTLFromContract(...) ?? details?.batchTTL` swap (matches the removed `fetchAuthoritativeBatchTTL` semantics, including Bee's negative-TTL normalisation); `isDefault` reference-equality against the filtered stamp list; per-batch worker-pool caching and utilization-broadcast routing; `batchID` hex normalisation and Zod validation; `enqueueWork` re-entrancy (no nesting, no wedged queue, no unhandled rejections); the demo selector and its fallback-to-Default; the `format:check` pathspec (all four tracked `CLAUDE.md` symlinks are at depth ≤ 1, so the excludes cover them).

### Test run

`pnpm --filter @snaha/swarm-id test` had 3 failures, all wall-clock/timing flakes rather than anything from this PR: the two known `roundtrip.test.ts` timeouts, plus `partition-lease.integration.test.ts:1715` ("solo-clean guard skip"), which asserts `elapsed < 750ms` and measured 1012ms. That one also failed when re-run in isolation on a loaded machine — worth confirming on a quiet one.

agazso:
Thanks for the review — all four findings were real; addressed in b78749e, 71e5bc1, c40c62d.

1. **teardown counter capture** (b78749e): the counters are now captured alongside the stampers, before the synchronous unbind. The test doubles were also hardened so this class of bug can't hide again: the stamper double only returns a counter while bound, and the `withAccountWriteLock` mock defers its callback a microtask like the real Web Locks API — with those two changes the existing teardown-flush test reproduced the bug before the fix.

2. **mid-join breaker reset** (b78749e): went with the suggested marker — `signalLeaseLost` records the lease session it fired for, and `ensureBatchJoined` re-checks it (plus `partitionLease === lease` and `disposed`) after `joinBatch` resolves, throwing `PartitionLeaseLostError` instead of binding. The marker is cleared in `finalizeDemote`, where the identity check takes over. New test covers the exact interleaving (signal lands while the join is in flight).

3. **adopt-path zero-seed** (71e5bc1): the snapshot now carries `priorHolderLeasedUntil` (the held partition's prior-holder `leasedUntil` from the acquire scan) through `serialize`/`hydrate`, so a re-adopted session's `joinBatch` scans the full lease span — and the `pointerInconclusive → readFailed` fail-safe arms too. `refreshFromSwarm` still clears and re-learns it on the next real scan. Didn't persist `unreadableLocks`: a partition we successfully claimed can't have come from an unreadable lock (acquire fails safe on those). New integration test: prior holder publishes >2 epochs back → cold acquire → serialize/hydrate/adopt → `joinBatch` resumes the exact counter.

4. **stale docs** (c40c62d): updated to `withAccountWriteLock` / `swarm-write-account-<accountId>`, including the cross-batch local exclusion.

Test run here: full `@snaha/swarm-id` suite green (1049/1049) on a standalone run, including `partition-lease.integration.test.ts:1715` ("solo-clean guard skip") — it also passed inside the full parallel suite on this (quiet) machine, so the 1012 ms you saw does look like load, not a regression. One `sequence/roundtrip.test.ts` timeout flake showed up in a later `check:all` run and passed in isolation, consistent with the known flakes.

vojtechsimetka (approved):

PR #539, opened 2026-08-14
docs: document the on-chain postage engine and the drive payment flow

vojtechsimetka:
**Stack 3/7** — base `stack/2-multichain` ([#532](https://github.com/snaha/swarm-id/pull/532)). Part of #309.

The two documents for the change that follows, split out of it so the implementation can be read *against* them rather than carrying them. 554 lines of Markdown, no code.

- **`docs/Postage-On-Chain-Engine.md`** — read this one first. The contract-facts table is what the ordering rule rests on: `increaseDepth` checks the post-dilution per-chunk balance against the ~24h floor **before** any compensation, so the compensating top-up must run first.
- **`docs/Drive-Payment-Flow.md`** — consumes the engine's API: how BZZ and xDAI reach the owner address without the external widget.

Both follow the house style of the existing `docs/` set (`Account-State.md`, `BatchWriteCoordinator.md`): present tense, "What it is" / "Position in the stack" / "Files and tests", with a `Known gaps` section instead of inline status markers. Acceptance criteria and per-section done/open bookkeeping live in the implementation PRs, not here.

### DIFF
--- a/docs/Drive-Payment-Flow.md
+++ b/docs/Drive-Payment-Flow.md
@@ -1,0 +1,280 @@
+# Drive payment flow — paying for storage in-app, from any chain
+
+Status: **design of record**; the implementation lands in
+[#533](https://github.com/snaha/swarm-id/pull/533) and
+[#534](https://github.com/snaha/swarm-id/pull/534). This describes how BZZ and xDAI reach the derived
+batch-owner address. What is then done with them on Gnosis is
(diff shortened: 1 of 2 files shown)


vojtechsimetka:
Superseded — the two documents now land inside #533, alongside the code they describe, so they never claim a status the tree does not have. Content is unchanged from the rewritten version reviewed here; the open items moved to #541, #542, #543 and #544.

PR #540, opened 2026-08-14
feat(dev): rebuild /dev around the chain, with a faucet and a real test drive

vojtechsimetka:
**Stack 3/6** — base `stack/2-multichain` ([#532](https://github.com/snaha/swarm-id/pull/532)). Part of #309. (The specs that used to sit between them, #539, are closed — they land with the code in [#533](https://github.com/snaha/swarm-id/pull/533).)

`/dev` was organised around a Bee node buying stamps for you, which stops being how any of this works once drives are paid for on-chain. It is now organised by what each tab talks to — and it lands **ahead** of that change, so the change itself can be tried by hand.

## Why this comes first

[#533](https://github.com/snaha/swarm-id/pull/533) removes the simulated settlement, so every purchase becomes a real on-chain buy. Without this PR there is no way for a person to fund one locally: no faucet, no dev script, and the widget only settles on mainnet. The e2e suites were fine — they fund out of band — but a reviewer clicking through had nothing.

## The Chain tab

- **Faucet** — any amount of xDAI or BZZ, to any address, from the float the chain snapshot bakes in. Prefilled with the selected account's postage signer.
- **Create drive to test with** — creates a real batch the account's **own postage signer owns on chain**, and attaches it as an ordinary drive. That replaces six copy-paste steps through Import batch, and it is the fixture `drive-onchain.test.ts` uses later in the stack.
- **Create expiring drive** — the same thing funded for a single day, so the product's "Expires soon" warning can be produced on purpose. The ordinary button funds 15 days, clear of the 7-day threshold: a fixture that is born flagged teaches everyone to read that warning as noise.
- A **`Connected to: Local | Production | Custom`** menu in the page's top right, switching both endpoints at once. It lives here rather than in the product's Network settings dialog, which ships; `Custom` names an endpoint that is neither preset, and editing that one opens the product dialog rather than a second copy of its fields.

## Which chain you are pointed at, said wherever it is asked

Three surfaces need the same answer, so all three now give it, from the same `chainIdentity` probe:

- the **`/dev` banner** — `GNOSIS MAINNET — these tools spend real funds` in red, or `Local dev chain, nothing here is real`;
- a **`Connected to: …` line under the RPC field in Network settings** — `Gnosis Chain`, `Gnosis Chain (fake) — funds here are not real`, or `Not reachable`. The URL is otherwise the only clue a user has, and a stale `localhost` looks exactly like the real thing. It probes the *saved* value, not the field being typed into, so editing does not fire a request per keystroke.

The Overview row for that endpoint is labelled plainly, `Gnosis Chain RPC:` — an earlier revision of this PR hardcoded `(fake)` there, which contradicted the MAINNET banner two lines above it the moment anyone pointed at the real chain. The banner asserts the identity; the row just names the endpoint.

## `$lib/payment/chain.ts` is new, and is why any of this can run here

Which chain an endpoint serves — **by genesis hash, not chain id**, since the local chain reports mainnet's 100 on purpose — and a client for it. The question is asked well before any operation exists: the faucet, the banner and the settings dialog all need it. `contract.ts` now resolves the PostageStamp address through it rather than by hostname, which is the same correction [#531](https://github.com/snaha/swarm-id/pull/531) made in the library.

It is extracted from `postage-onchain.ts` in [#533](https://github.com/snaha/swarm-id/pull/533), which keeps the operations and imports the chain resolution.

### An unproven genesis must not read as "fake"

Review caught the first cut failing open. `probeGenesisHash` returned `data.result?.hash`, so an endpoint answering a JSON-RPC `error` — a rate-limited or pruned Gnosis node — yielded `undefined`, which is not the mainnet genesis, which reads as *not mainnet*. A stub answering `eth_chainId: 0x64` and refusing `eth_getBlockByNumber` was duly labelled `Local dev chain, nothing here is real`, under which the faucet and **Create owned batch** spend real money.

Both probes now go through one checked helper that rejects on non-2xx, on a JSON-RPC `error`, and on a missing result. An endpoint that cannot prove its genesis is `No chain reachable` — loud, and in the safe direction. (This also answers Copilot's note about unchecked HTTP/JSON-RPC status on `probeChainId`; its other suggestion, replacing `AbortSignal.timeout()`, I left alone — it is the established pattern in `multichain/src/fetch.ts` and its browser support is a strictly older baseline than the passkeys this app already requires.)

### The banner could report a healthy endpoint dead

`{#await chainIdentity(...)}` was unkeyed, so switching away from an unreachable RPC kept rendering the failed branch — with the *new* url interpolated into it, since the message reads the url from the store rather than the promise. Selecting a working endpoint showed `No chain reachable at <it>` instantly, no probe involved. The block is now keyed on the endpoint.

## The mock stamp purchase stays, for now

Moved onto the Chain tab, not deleted. It is still the only way the product Add-drive flow completes off mainnet, and `mockStampEnabled` defaults to `import.meta.env.DEV` — **on** — so removing the toggles here would strand anyone who has it enabled with no way to turn it off. It goes with the flow it fakes, in [#533](https://github.com/snaha/swarm-id/pull/533).

## Two more things review turned up

- **The node has not seen the batch yet.** A batch bought straight from the contract exists on chain before Bee's chain sync reaches it, and stamping with it until then is rejected `400 invalid batch id` — so attaching the drive immediately made the next account sync fail for no visible reason. `createTestDrive` now waits for `/status` to pass the batch's block, best effort.
- **The mock ignored the size you picked.** It settled a fabricated batch at a hardcoded depth 20, so a mocked Add drive at 97.7 MB produced a 600 MB drive. The real widget chooses the depth itself, which is why no plain `depth` option exists; a mock-only `mockDepth` now carries the selection through.

## Verification

`pnpm check:all` green. Exercised by hand against a `--fresh --pull` bee-compose 0.2.0 cluster:

| Step | Result |
| --- | --- |
| Banner, public RPC | "GNOSIS MAINNET — these tools spend real funds" |
| Banner, local chain | "Local dev chain, nothing here is real" — correct despite chain id 100 |
| Settings, public RPC | "Connected to: Gnosis Chain" |
| Settings, local chain | "Connected to: Gnosis Chain (fake) — funds here are not real" — chain id 100, genesis says otherwise |
| Settings, nothing running | "Connected to: Not reachable", in the destructive tone |
| PostageStamp resolution | mainnet address `0x45a15023…` on localhost |
| Faucet | signer 0 → 0.05 xDAI / 1 BZZ; faucet 100 → 99.9499, 250 → 249 |
| Create drive to test with | batch `f91fd63d…`, depth 20 |
| On-chain owner | `0xe21393b4…58f` — the account's derived postage signer |
| Product UI | an ordinary drive, 600 MB |
| Drive lifetimes | 15.00 d default → `15 days left`; 1.00 d expiring → `Expires soon` |
| Genesis stub (chain id 100, genesis refused) | `No chain reachable` — was `Local dev chain, nothing here is real` |
| Dead RPC → Use local | reports the live chain; was stuck on the failure branch |
| Endpoints, production | `Cluster worker` row hidden; no dead `localhost:16331` |
| Mocked Add drive at 97.7 MB | a 97.7 MB drive |
| Batch visible to node | SOC upload accepted after the wait (`ed32ed15…`) |

A SOC upload stamped with the on-chain batch was accepted, so the drive is genuinely usable rather than merely listed.

Pre-existing `rejectAfter`/`Promise.race` timeout leaks found while reviewing this are filed separately as #546 — none of them are touched here.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -229,6 +229,9 @@ importers:
       '@snaha/swarm-id':
         specifier: workspace:*
         version: link:../lib
+      '@swarm-id/multichain':
+        specifier: workspace:*
+        version: link:../multichain
       '@web3-onboard/core':
         specifier: ^2.24.1
         version: 2.24.1(typescript@5.8.3)
@@ -238,6 +241,9 @@ importers:
       ethers:
         specifier: 6.15.0
         version: 6.15.0
+      viem:
+        specifier: ^2.55.10
+        version: 2.55.13(typescript@5.8.3)(zod@4.4.3)
     devDependencies:
       '@eslint/js':
         specifier: 9.39.1
--- a/ui/knip.ts
+++ b/ui/knip.ts
@@ -11,7 +11,7 @@ const config: KnipConfig = {
   },
   // Knip cannot resolve the workspace package through its exports map, so it
   // would misreport the dependency as unused.
-  ignoreDependencies: ['@snaha/swarm-id'],
+  ignoreDependencies: ['@snaha/swarm-id', '@swarm-id/multichain'],
   ignoreExportsUsedInFile: true,
 }
 
--- a/ui/package.json
+++ b/ui/package.json
@@ -53,8 +53,10 @@
   "dependencies": {
     "@ethersphere/bee-js": "^11.1.1",
     "@snaha/swarm-id": "workspace:*",
+    "@swarm-id/multichain": "workspace:*",
     "@web3-onboard/core": "^2.24.1",
     "@web3-onboard/injected-wallets": "^2.11.3",
-    "ethers": "6.15.0"
+    "ethers": "6.15.0",
+    "viem": "^2.55.10"
   }
 }
--- a/ui/src/lib/components/drive-add-dialog.svelte
+++ b/ui/src/lib/components/drive-add-dialog.svelte
@@ -181,6 +181,8 @@
         mocked: devSettingsStore.data.mockStampEnabled,
         mockPopup: devSettingsStore.data.mockStampPopup,
         mockError: devSettingsStore.data.mockStampResult === 'error',
+        // Only the mock can honour this — the real widget picks the size itself.
+        mockDepth: depthValue === '' ? undefined : Number(depthValue),
         onSuccess: (batch) => {
           if (!attempt.current) {
             return
--- a/ui/src/lib/components/network-settings-dialog.svelte
+++ b/ui/src/lib/components/network-settings-dialog.svelte
@@ -9,22 +9,12 @@
   import { Button } from '$lib/components/ui/button'
   import { Dialog } from '$lib/components/ui/dialog'
   import { Input } from '$lib/components/ui/input'
-  import { chainIdentity, probeChainId } from '$lib/payment/chain'
+  import { chainIdentity } from '$lib/payment/chain'
   import { networkSettingsStore } from '$lib/stores/network-settings.svelte'
 
   interface Props {
     onclose: () => void
   }
-
-  /**
-   * The local endpoints "Use local" fills in. Two chains can be serving Gnosis
-   * locally — bee-compose's cluster and the standalone snapshot — so the button
-   * asks which is actually answering rather than making the user remember. The
-   * cluster comes first: it also serves the Bee node, so if it is up that is
-   * the environment you want.
-   */
-  const LOCAL_BEE_NODE_URL = 'http://localhost:1633/'
-  const LOCAL_GNOSIS_RPC_URLS = ['http://localhost:9545', 'http://localhost:8545']
 
   let { onclose }: Props = $props()
 
@@ -70,24 +60,6 @@
     beeNodeUrl = DEFAULT_BEE_NODE_URL
     gnosisRpcUrl = DEFAULT_GNOSIS_RPC_URL
   }
-
-  /**
-   * Point at whichever local chain is running. Dev builds only — the button is
-   * not rendered otherwise, and typing two localhost URLs by hand is the step
-   * that most often goes wrong when setting the environment up.
-   */
-  async function useLocal() {
-    beeNodeUrl = LOCAL_BEE_NODE_URL
-    const reachable = await Promise.all(
-      LOCAL_GNOSIS_RPC_URLS.map((url) =>
-        probeChainId(url).then(
-          () => url,
-          () => undefined,
-        ),
-      ),
-    )
-    gnosisRpcUrl = reachable.find((url) => url !== undefined) ?? LOCAL_GNOSIS_RPC_URLS[0]
-  }
 </script>
 
 <Dialog title="Network settings" {onclose}>
@@ -126,11 +98,6 @@
   <div class="flex w-full items-center gap-2">
     <Button disabled={!canSave} onclick={save}>Save</Button>
     <Button variant="outline" onclick={onclose}>Cancel</Button>
-    {#if import.meta.env.DEV}
-      <Button variant="ghost" class="ml-auto" onclick={useLocal}>Use local</Button>
-      <Button variant="ghost" onclick={resetToDefaults}>Reset</Button>
-    {:else}
-      <Button variant="ghost" class="ml-auto" onclick={resetToDefaults}>Reset</Button>
-    {/if}
+    <Button variant="ghost" class="ml-auto" onclick={resetToDefaults}>Reset</Button>
   </div>
 </Dialog>
--- a/ui/src/lib/components/network-settings-dialog.svelte
+++ b/ui/src/lib/components/network-settings-dialog.svelte
@@ -9,11 +9,22 @@
   import { Button } from '$lib/components/ui/button'
   import { Dialog } from '$lib/components/ui/dialog'
   import { Input } from '$lib/components/ui/input'
+  import { chainIdentity, probeChainId } from '$lib/payment/chain'
   import { networkSettingsStore } from '$lib/stores/network-settings.svelte'
 
   interface Props {
     onclose: () => void
   }
+
+  /**
+   * The local endpoints "Use local" fills in. Two chains can be serving Gnosis
+   * locally — bee-compose's cluster and the standalone snapshot — so the button
+   * asks which is actually answering rather than making the user remember. The
+   * cluster comes first: it also serves the Bee node, so if it is up that is
+   * the environment you want.
+   */
+  const LOCAL_BEE_NODE_URL = 'http://localhost:1633/'
+  const LOCAL_GNOSIS_RPC_URLS = ['http://localhost:9545', 'http://localhost:8545']
 
   let { onclose }: Props = $props()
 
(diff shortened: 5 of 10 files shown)


agazso:
Critical review — findings, most interesting first:

**1. The Import section's `resolvedContract` effect is dead code with a race — and it probes per keystroke.**
`settingsFor` in `chain.ts` throws for any chain id other than Gnosis and otherwise always returns `gnosisMainnetSettings(...)` with only `rpcUrls` overridden — so `(await postageChain(url)).settings.addresses.postageStamp` can never be anything but `DEFAULT_POSTAGE_STAMP`. The `$effect` on `importRpcUrl` therefore fires two RPC probes (chainId + genesis) on every keystroke into the RPC URL field — the exact behavior the settings-dialog comment deliberately avoids — has a last-write-wins race between in-flight probes, and can only ever compute the constant it falls back to. Replace the whole thing with `const resolvedContract = DEFAULT_POSTAGE_STAMP`. Relatedly, the kept hint text *"local RPC → local deployment, else Gnosis mainnet"* is now false by the PR's own premise (there is no local-only deployment); same for the `contract.ts` comment framing — the real effect of routing through `postageChain` there is endpoint validation + a new failure mode, not address resolution.

**2. Settings dialog: "Connected to:" describes the saved endpoint but renders under the edited field.**
Type the real mainnet RPC over a stale `localhost` value and, until you hit Save, the dialog shows *"Gnosis Chain (fake) — funds here are not real"* directly beneath the mainnet URL you just entered. Probing only the saved value is deliberate (fine), but the rendering isn't honest about it. One-line fix: hide the line when `gnosisRpcUrl.trim() !== networkSettingsStore.gnosisRpcUrl`. This dialog ships to production and the string is money-safety wording, so worth fixing.

**3. No structural mainnet guard in `chain-funding.ts`.**
The faucet and `createOwnedBatchOnChain` are only safe against real mainnet by accident: the faucet key holds nothing there, and `anvil_setBalance` fails on a real node. Both facts are true today but live two packages away. Since `chainIdentity()` is already computed on this path, `if ((await chainIdentity()).isMainnet) throw` at the top of `sendFromFaucet`/`createOwnedBatchOnChain` turns accidental safety into designed safety for two lines. The red banner warns humans; this would stop the code.

**4. The banner can't recover without a reload.**
`{#key url}{#await chainIdentity(url)}` retries only when the URL changes. Start the page before the local chain is up and the banner stays "No chain reachable" forever, even though the failed probe was correctly evicted from the cache — there's just nothing to re-trigger it. Dev-page-only, navigating away and back fixes it; a retry key or button would be a small win.

**5. `endpointMode` is an exact-string match.**
Saving `http://localhost:1633` (no trailing slash) through the settings dialog shows "Custom" for a genuinely local setup, because `LOCAL_BEE_NODE_URL` carries the slash. Cosmetic, dev-only; normalizing via `new URL(url).href` before comparing would cover it.

**6. Zero tests for `chain.ts`.**
The module's two safety invariants — an unproven genesis rejects rather than reading as "not mainnet", and a dev chain's settings *never* include public RPC fallbacks — are exactly the kind of thing a later refactor silently breaks, and both are trivially unit-testable (mock `fetch`). The identity-cache eviction logic likewise.

**Nits:** `viem` becomes a direct prod dependency of `ui/` for `formatUnits`/`parseUnits`/`generatePrivateKey`, all of which `ethers` (already a dep) provides — defensible since `multichain` is viem-based, but worth a conscious decision. `createTestDrive` derives the postage signer twice (once inside `createOwnedBatchOnChain`, once for the key) — harmless, deterministic.

Checked and found sound: the fail-closed `jsonRpc` helper and its use in both probes; the identity/client cache eviction; the out-of-order guard in `refreshFunds`; `mockDepth` (fed from a fixed-option Select, so `Number()` is safe); the `{#key}` fix for the banner's stale-failure bug; and the receipt/block-wait ordering in `createTestDrive` (`createBatch` polls its own receipt, wait happens before `addStamp`, and Bee's `/status` really serves `lastSyncedBlock`).

vojtechsimetka:
The page should show warning if I am on localhost and not connected to local. I just spend some real DAI and BZZ.

agazso (approved):
There are three code-level issues that I found:

1. chain.ts:93 — fourth hand-rolled JSON-RPC POST transport in the monorepo (lib postJsonRpc, multichain fetch.ts, dev-rpc.ts, now this). Each copy has different timeout/guard behavior. Cheaper to export lib's postJsonRpc and layer the null guard on it.
2. chain-funding.ts:268 — hand-rolled Bee /status fetch with an untyped cast, while bee-js (already imported in this file) has getStatus() with strict parsing; a field rename silently defeats the poller.
3. chain.ts:240 — prefix0x(address) as `\0x${string}` string-juggling for an owner address, against ui/AGENTS.md's explicit `new EthAddress(value)` rule; sibling code in the same feature already does it right.

vojtechsimetka:
Thanks @agazso — two of the three are in [`1b08f5ec`](https://github.com/snaha/swarm-id/commit/1b08f5ec0), the third is #552.

**2 — `chain-funding.ts` `/status`** and **3 — `chain.ts` owner address**: both fixed. They turned out to be the same mistake twice: trusting a string where a type already existed for it.

`ownerFunds` now parses with `new EthAddress(address).toChecksum()`. You were right that this is not just style — prefixing cannot fail, so a malformed address travelled to the RPC and came back as a balance of zero, which on the page reads as a funded account that has spent everything rather than as a typo. `chain.test.ts` gained cases for it; four of them fail against the old line, including a lower-case address, which used to be passed through unchecksummed.

`waitForNodeToSeeBatch` reads `bee.getStatus()`. Both stop conditions survive: a `BeeResponseError` is an endpoint that answered and refused — a gateway — so stop asking, while a transport failure is a node coming back up, so keep waiting to the capped run. The cast is gone, so a renamed `lastSyncedBlock` is now a type error rather than a poller that silently never advances.

**1 — the fourth JSON-RPC transport** is real, and I filed it as #552 rather than doing it here. The narrow form of it — export `postJsonRpc`, layer the null guard — does not retire a transport: `postJsonRpc` has no timeout, so `chain.ts` would either lose its 5 s probe deadline or push one into a published package on behalf of two unrelated callers, and since `postJsonRpc` deliberately returns the raw envelope the guards all stay where they are. Four transports would become four transports with an import between them, plus a new public export on `@snaha/swarm-id`, in a PR about `/dev` tooling. The issue has the table of how the four currently disagree and what unifying them would have to preserve.

`pnpm check:all` green, rebased onto main.

vojtechsimetka:
Correction to the note above, and to [`1b08f5ec`](https://github.com/snaha/swarm-id/commit/1b08f5ec0) — now [`fb830b82`](https://github.com/snaha/swarm-id/commit/fb830b82c).

I wrote that a `BeeResponseError` is "an endpoint that answered and refused". It is not: bee-js wraps **every** `AxiosError` in one, transport failures included ([`utils/http.js:71`](https://github.com/ethersphere/bee-js/blob/master/src/utils/http.ts)). A refused connection — a node coming back up, the very case this wait exists for — arrives as a `BeeResponseError` with no `status`, so keying the stop on the error type would have made `waitForNodeToSeeBatch` return immediately in exactly the situation it was written for, and the drive would be attached before the node could stamp with it again.

The stop is now on `error.status !== undefined`, which is the real question — did something answer — and matches what the hand-rolled version's `!response.ok` was checking.

